Issue #7, opened 2019-01-17
Type conversion builtin functions to take default values


d5:
For example, 

```golang
v1 := int("foo")     // v1 == undefined
v2 := int("foo", 10) // v2 == 10
```

mdbazuin:
I've been messing around a bit trying to understand the code, so i tried messing around with the type conversion default.

This is basically a proof of concept, would something like this work for you? if so i'll make a pr for all built in conversion types including tests.

If needed i can also add an error type ErrWrongObjectType or something like that when the default isn't an Int (for this specific case)

```go
func builtinInt(args ...Object) (Object, error) {
	argsLen := len(args)
	if !(argsLen == 1 || argsLen == 2) {
		return nil, ErrWrongNumArguments
	}

	switch arg := args[0].(type) {
	case *Int:
		return arg, nil
	case *Float:
		return &Int{Value: int64(arg.Value)}, nil
	case *Char:
		return &Int{Value: int64(arg.Value)}, nil
	case *Bool:
		if arg.Value {
			return &Int{Value: 1}, nil
		}
		return &Int{Value: 0}, nil
	case *String:
		n, err := strconv.ParseInt(arg.Value, 10, 64)
		if err == nil {
			return &Int{Value: n}, nil
		}
	}

	if argsLen == 2 {
		switch arg := args[1].(type) {
		case *Int:
			return arg, nil
		}
	}

	return UndefinedValue, nil
}
```

d5:
Yeah, I'd appreciate if you could create a PR for this! 😄 Btw, I just merged a big PR that has some code changes around these type conversions. But it shouldn't be too different.

d5:
I also added a documentation on types: https://github.com/d5/tengo/wiki/Variable-Types. Just FYI.

mdbazuin:
I will create a PR shortly, should be able to adjust it to the changes you made in the latest PR in like an an hour or so

PR #14, opened 2019-01-18
Type conversion builtin take default values

mdbazuin:
#7

objects/builtin_convert.go:
- Added defaults to all builtin convert functions except
bool as this one returns false for undefined, instead of undefined.

runtime/vm_builtin_test.go: 
- Added builtin bytes convert test
- Added tests for checking default value
behaviour for builtin convert funcs

### DIFF
--- a/objects/builtin_convert.go
+++ b/objects/builtin_convert.go
@@ -16,9 +16,7 @@ func builtinString(args ...Object) (Object, error) {
 	}
 
 	if argsLen == 2 {
-		if _, ok := args[1].(*String); ok {
-			return args[1], nil
-		}
+		return args[1], nil
 	}
 
 	return UndefinedValue, nil
@@ -40,9 +38,7 @@ func builtinInt(args ...Object) (Object, error) {
 	}
 
 	if argsLen == 2 {
-		if _, ok := args[1].(*Int); ok {
-			return args[1], nil
-		}
+		return args[1]
 	}
 
 	return UndefinedValue, nil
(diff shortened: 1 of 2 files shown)


d5:
This is great! It's super clean.

One thing that I'm trying to figure out is whether the second argument Object should be returned as-is with no type checking/conversion. I'm not sure if it's good or bad, but, maybe it can enable something interesting. 

```golang
x = int("not a number", x) // do not modify 'x' if failed

b := true
y = float(b, int(b)) // do int() if float() fails
```

Also it will simply translate `int(x)` as equivalent expression to `int(x, undefined)`.

mdbazuin:
That's a great suggestion. I think it would definitely enable interesting behaviour. I think it isn't really default behaviour anymore, but an or functionality. Sounds good to me, do you want me to implement that?

d5:
Lol, yeah. Honestly I don't know if this is really a good idea or not, but, I'm open to experimental approaches.

mdbazuin:
I think it does enable some nice error handling if you want, since you can just say if it failed give me an error

d5:
@mdbazuin actually test fails. Can you fix compile errors?
```
objects/builtin_convert.go:41:3: not enough arguments to return
	have (Object)
	want (Object, error)
objects/builtin_convert.go:63:3: not enough arguments to return
	have (Object)
	want (Object, error)
objects/builtin_convert.go:102:3: not enough arguments to return
	have (Object)
	want (Object, error)
objects/builtin_convert.go:125:3: not enough arguments to return
	have (Object)
	want (Object, error)
```

d5:
@mdbazuin never mind, just fixed it.

mdbazuin:
I had that fixed locally, only forgot to add it to the commit after i added the updated tests

d5:
It's alright. Super minor.

PR #29, opened 2019-01-21
Added builtin functions to_json and from_json

mdbazuin:
Fixes bug in script/conversion.go:interfaceToObject:
- Bug was caused by initialising var `arr` with length `len(v)`
- Followed by setting the `interfaceToObject(e)` in an append to `arr`
- This causes a slice for `len(v) = 3` as `[nil, nil, nil, 1, 2, 3,]`

Added conversion function ToMap:
- Takes an object and tries to make it a `map[string]interface{}`

Added conversion function ToSlice:
- Takes an object and tries to make it a `[]interface{}`

Added builtin function `to_json`:
- Takes an `array` or `map` as argument
- Returns value as `bytes`

Added builtin function `from_json`: 
- Takes a `bytes` or `string` as argument
- Returns value as `map`

### DIFF
--- a/objects/builtin_json.go
+++ b/objects/builtin_json.go
@@ -16,3 +16,33 @@ func builtinToJSON(args ...Object) (Object, error) {
 
 	return &Bytes{Value: res}, nil
 }
+
+func builtinFromJSON(args ...Object) (Object, error) {
+	if len(args) != 1 {
+		return nil, ErrWrongNumArguments
+	}
+
+	var target interface{}
+
+	switch o := args[0].(type) {
+	case *Bytes:
+		err := json.Unmarshal(o.Value, &target)
+		if err != nil {
+			return nil, err
+		}
+	case *String:
+		err := json.Unmarshal([]byte(o.Value), &target)
+		if err != nil {
+			return nil, err
+		}
+	default:
+		return nil, ErrInvalidTypeConversion
+	}
+
+	res, err := FromInterface(target)
+	if err != nil {
+		return nil, err
+	}
+
+	return res, nil
+}
--- a/objects/builtin_json.go
+++ b/objects/builtin_json.go
@@ -1,0 +1,18 @@
+package objects
+
+import (
+	"encoding/json"
+)
+
+func builtinToJSON(args ...Object) (Object, error) {
+	if len(args) != 1 {
+		return nil, ErrWrongNumArguments
+	}
+
+	res, err := json.Marshal(objectToInterface(args[0]))
+	if err != nil {
+		return nil, err
+	}
+
+	return &Bytes{Value: res}, nil
+}
--- a/objects/conversion.go
+++ b/objects/conversion.go
@@ -1,6 +1,7 @@
 package objects
 
 import (
+	"fmt"
 	"strconv"
 )
 
@@ -137,4 +138,59 @@ func ToByteSlice(o Object) (v []byte, ok bool) {
 
 	//ok = false
 	return
+}
+
+// FromInterface will attempt to convert an interface{} v to a Tengo Object
+func FromInterface(v interface{}) (Object, error) {
+	switch v := v.(type) {
+	case nil:
+		return &Undefined{}, nil
+	case string:
+		return &String{Value: v}, nil
+	case int64:
+		return &Int{Value: v}, nil
+	case int:
+		return &Int{Value: int64(v)}, nil
+	case bool:
+		return &Bool{Value: v}, nil
+	case rune:
+		return &Char{Value: v}, nil
+	case byte:
+		return &Char{Value: rune(v)}, nil
+	case float64:
+		return &Float{Value: v}, nil
+	case []byte:
+		return &Bytes{Value: v}, nil
+	case error:
+		return &Error{Value: &String{Value: v.Error()}}, nil
+	case map[string]Object:
+		return &Map{Value: v}, nil
+	case map[string]interface{}:
+		kv := make(map[string]Object)
+		for vk, vv := range v {
+			vo, err := FromInterface(vv)
+			if err != nil {
+				return nil, err
+			}
+			kv[vk] = vo
+		}
+		return &Map{Value: kv}, nil
+	case []Object:
+		return &Array{Value: v}, nil
+	case []interface{}:
+		arr := make([]Object, len(v), len(v))
+		for i, e := range v {
+			vo, err := FromInterface(e)
+			if err != nil {
+				return nil, err
+			}
+
+			arr[i] = vo
+		}
+		return &Array{Value: arr}, nil
+	case Object:
+		return v, nil
+	}
+
+	return nil, fmt.Errorf("unsupported value type: %T", v)
 }
--- a/objects/conversion.go
+++ b/objects/conversion.go
@@ -140,6 +140,38 @@ func ToByteSlice(o Object) (v []byte, ok bool) {
 	return
 }
 
+// objectToInterface attempts to convert an object o to an interface{} value
+func objectToInterface(o Object) (res interface{}) {
+	switch o := o.(type) {
+	case *Int:
+		res = o.Value
+	case *String:
+		res = o.Value
+	case *Float:
+		res = o.Value
+	case *Bool:
+		res = o.Value
+	case *Char:
+		res = o.Value
+	case *Bytes:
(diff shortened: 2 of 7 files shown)


objects/conversion.go
```diff
@@ -138,3 +139,123 @@ func ToByteSlice(o Object) (v []byte, ok bool) {
 	//ok = false
 	return
 }
+
+// ToMap will try to convert object o to map[string]interface value.
+func ToMap(o Object) (v map[string]interface{}, ok bool) {
+	v = make(map[string]interface{})
+	switch o := o.(type) {
+	case *Map:
```
d5: `v` should be assigned (making map) inside here.

script/conversion.go
```diff
@@ -70,13 +70,13 @@ func interfaceToObject(v interface{}) (objects.Object, error) {
...
 		arr := make([]objects.Object, len(v), len(v))
-		for _, e := range v {
+		for i, e := range v {
 			vo, err := interfaceToObject(e)
 			if err != nil {
 				return nil, err
 			}
 
-			arr = append(arr, vo)
+			arr[i] = vo
```
d5: good catch. thanks

objects/conversion.go
```diff
@@ -138,3 +139,123 @@ func ToByteSlice(o Object) (v []byte, ok bool) {
...
+
+	//ok = false
+	return
+}
+
+// objectToInterface attempts to convert an object o to an interface{} value
+func objectToInterface(o Object) (res interface{}) {
+	switch o := o.(type) {
+	case *Int:
+		res, _ = ToInt(o)
```
d5: Should be `res = o.Value` as `o` is already known to be `*Int`. Same for other primitive type cases here.

objects/conversion.go
```diff
@@ -138,3 +139,123 @@ func ToByteSlice(o Object) (v []byte, ok bool) {
...
+		res, _ = ToFloat64(o)
+	case *Bool:
+		res, _ = ToBool(o)
+	case *Char:
+		res, _ = ToRune(o)
+	case *Bytes:
+		res, _ = ToByteSlice(o)
+	case *Array:
+		values := o.Value
+		resSlice := make([]interface{}, len(values))
```
d5: I don't see why separate `values` and `resSlice` had to be created.

objects/conversion.go
```diff
@@ -138,3 +139,123 @@ func ToByteSlice(o Object) (v []byte, ok bool) {
...
+		for i, v := range values {
+			nestedRes := objectToInterface(v)
+			resSlice[i] = nestedRes
```
d5: also this can be `resSlice[I] = objectToInterface(v)`?

objects/conversion.go
```diff
@@ -138,3 +139,123 @@ func ToByteSlice(o Object) (v []byte, ok bool) {
...
+		}
+		res = resSlice
+	case *Map:
+		resMap := make(map[string]interface{})
+		for key, v := range o.Value {
+			nestedRes := objectToInterface(v)
+			resMap[key] = nestedRes
+		}
+		res = resMap
```
d5: Same as *Array: no need to create a separate variable `resMap` and line #198-199 can be a one-line.

objects/conversion.go
```diff
@@ -138,3 +139,123 @@ func ToByteSlice(o Object) (v []byte, ok bool) {
...
+	case Object:
+		return o
+	}
+	return res
+}
+
+// interfaceToObject will attempt to convert an interface{} v to a Tengo Object. Copy from script/conversion.go
+func interfaceToObject(v interface{}) (Object, error) {
```
d5: If this is an exact copy from `script/conversion.go`, let's make this a public function, like `objects.FromInterface(interface{})` and have script-side code to call this function.

objects/builtin_json.go
```diff
@@ -0,0 +1,66 @@
...
+		if !ok {
+			return nil, ErrInvalidTypeConversion
+		}
+	case *Map:
+		toMarshal, ok = ToMap(o)
+		if !ok {
+			return nil, ErrInvalidTypeConversion
+		}
+	default:
+		return nil, errors.New("to_json requires an array or map as argument")
```
d5: It does not have to be an array or a map. Primitive values are also valid JSON. https://play.golang.org/p/H7u25AhFsVZ

objects/builtin_json.go
```diff
@@ -0,0 +1,66 @@
...
+		if err != nil {
+			return nil, err
+		}
+	case *String:
+		err := json.Unmarshal([]byte(o.Value), &target)
+		if err != nil {
+			return nil, err
+		}
+	default:
+		return nil, errors.New("marshal only takes an array or map as parameter")
```
d5: Let's just use `ErrInvalidTypeConversion` error.

mdbazuin:
Thanks for the feedback!

I'll fix it when i get home

objects/conversion.go
mdbazuin: Should be fixed

objects/builtin_json.go
```diff
@@ -0,0 +1,48 @@
...
+
+	return &Bytes{Value: res}, nil
+}
+
+func builtinFromJSON(args ...Object) (Object, error) {
+	if len(args) != 1 {
+		return nil, ErrWrongNumArguments
+	}
+
+	target := make(map[string]interface{})
```
wangtuanjie: maybe json.RawMessage?

runtime/vm_builtin_test.go
```diff
@@ -113,4 +113,31 @@ func TestBuiltinFunction(t *testing.T) {
...
+	expect(t, `out = from_json("{\"foo\":\"bar\"}").foo`, "bar")
+	expect(t, `out = from_json("{\"foo\":1.8}").foo`, 1.8)
+	expect(t, `out = from_json("{\"foo\":true}").foo`, true)
+	expect(t, `out = from_json("{\"foo\":[\"bar\",1,1.8,56,true]}").foo[0]`, "bar")
+	expect(t, `out = from_json("{\"foo\":[[\"bar\",1],[\"bar\",1]]}").foo[0][0]`, "bar")
+	expect(t, `out = from_json("{\"foo\":{\"bool\":true,\"char\":56,\"float\":1.8,\"int\":1,\"string\":\"bar\"}}").foo.bool`, true)
+	expect(t, `out = from_json("{\"foo\":{\"map1\":{\"string\":\"bar\"},\"map2\":{\"int\":\"1\"}}}").foo.map1.string`, "bar")
+
+	expect(t, `out = from_json("5")`, 5.0)
+	expect(t, `out = from_json("[\"bar\",1,1.8,56,true]")[0]`, "bar")
```
d5: Just so you know, using array as expected value is totally fine: e.g.  `expect(t, ...., ARR{"bar", 1, 1.8, 56, true})`

objects/conversion.go
```diff
@@ -138,3 +139,118 @@ func ToByteSlice(o Object) (v []byte, ok bool) {
...
+		}
+		ok = true
+	}
+
+	//ok = false
+	return
+}
+
+// ToSlice will try to convert object o to []interface{} value.
+func ToSlice(o Object) (v []interface{}, ok bool) {
```
d5: Let's remove it if not used.

objects/conversion.go
```diff
@@ -138,3 +139,118 @@ func ToByteSlice(o Object) (v []byte, ok bool) {
 	//ok = false
 	return
 }
+
+// ToMap will try to convert object o to map[string]interface value.
+func ToMap(o Object) (v map[string]interface{}, ok bool) {
```
d5: Let's remove it if not used.

runtime/vm_builtin_test.go
mdbazuin: I added the ARR{"bar", 1, 1.8, 56, true} as expected value

objects/conversion.go
mdbazuin: Removed this whole commit, includes ToSlice, ToMap and objectToInterface. The latter was only being used by ToSlice and ToMap

objects/conversion.go
mdbazuin: Removed this whole commit, includes ToSlice, ToMap and objectToInterface. The latter was only being used by ToSlice and ToMap

PR #47, opened 2019-01-25
build: add "make build" command

Chyroc:


d5:
@Chyroc thanks for the PR. Unfortunately, I cannot merge this because I have different plans for `build` command.

PR #61, opened 2019-01-30
Fix possible typo

MrH100000:
It looks like there is a typo in one of the comments about compiled bytecode in the documentation.


d5:
Sorry, this will be fixed in another larger PR. But, thanks for letting me know!!

PR #79, opened 2019-02-05
Added nil-coalescing operator.

earncef:
Added a nil-coalescing operator similar to one found in Swift and PHP.

### DIFF
--- a/compiler/ast/valid_expr.go
+++ b/compiler/ast/valid_expr.go
@@ -1,0 +1,29 @@
+package ast
+
+import (
+	"github.com/d5/tengo/compiler/source"
+)
+
+// ValidExpr represents a ternary conditional expression.
+type ValidExpr struct {
+	Cond        Expr
+	True        Expr
+	False       Expr
+	QuestionPos source.Pos
+}
+
+func (e *ValidExpr) exprNode() {}
+
+// Pos returns the position of first character belonging to the node.
+func (e *ValidExpr) Pos() source.Pos {
+	return e.Cond.Pos()
+}
+
+// End returns the position of first character immediately after the node.
+func (e *ValidExpr) End() source.Pos {
+	return e.False.End()
+}
+
+func (e *ValidExpr) String() string {
+	return "(" + e.Cond.String() + " ?? " + e.False.String() + ")"
+}
--- a/compiler/compiler.go
+++ b/compiler/compiler.go
@@ -570,6 +570,33 @@ func (c *Compiler) Compile(node ast.Node) error {
 		// update second jump offset
 		curPos = len(c.currentInstructions())
 		c.changeOperand(jumpPos2, curPos)
+
+	case *ast.ValidExpr:
+		if err := c.Compile(node.Cond); err != nil {
+			return err
+		}
+
+		// first jump placeholder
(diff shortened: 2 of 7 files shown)


d5:
Thanks for the PR.

Basically we can achieve the same thing using `||` operator.

```golang
a := 10 || 20
b := 0 || 10
```

I thought about this when adding the ternary condition expression, which is _almost_ same as using `&& ||` operators.

```golang
a := 10 && 20 || 30
b := 0 && foo() || bar()
c := 0 ? foo() : bar()
```

But the difference between ternary conditional and `&& ||` was that the latter ends up evaluating both true expression and false expression if the true expression evaluates to false (e.g. `foo()` returns a falsy value), while `?:` only evaluates either true expression or false expression.

But for the case of `||` operator, I'm not sure if there's any differences. But maybe I'm missing something?

earncef:
Sorry, I think the PR is not complete. 

I'll need to put in more work, but the point of the operator I am adding is to deal with expanding arrays and maps, in addition to what can be done with the ternary expression, something that can't be done by the || operator. Something like the below should result with b = 5 instead of an index out of bounds exception in both cases:
```
a := []
b:= a[1] ?? 5
b:= a[1][1] ?? 5
```
Are you willing to add the operator as described above if I finish the PR?

d5:
I see. Btw, I'm not against the idea of adding new operators. I just want to make sure it's not redundant. 

When looking at your example, the issue here is that the array indexer raises a "run-time error" for out-of-bounds index (instead of `undefined` value like maps). My reasoning behind this run-time behavior was because the array does not allow setting elements at index of out of bounds. e.g. `a := []; a[1] = 5 // expand the array` while map allows such index-set. So, for now, I think this is a different problem, unless we decide to change this array indexer behavior _(which I'm totally open for further discussion)_.

In the mean time, maybe we can add a builtin function, something like:

```golang
array_get(a, 1) // returns 'undefined' instead of raising run-time error
```

_(There should be better name than `array_get` though.)_

Then, we can add index-get operator to `Undefined` type (to simply return `undefined`). That way we can do:

```golang
a := []
b := array_get(a, 1) || 5
c := array_get(a, array_get(a, 1)) || 5

// works for maps too
d := {}
e := d.foo.bar // 'd.foo' -> undefined, undefined.bar -> undefined
```

earncef:
Ah, that makes sense. I didn't notice the Undefined type before you mentioned it. 

It might make life much easier without a redundant operator if an array would return Undefined within IndexGet. While I see your reasoning for raising a runtime error, the issue with the builtin function would be that it will make the code very unreadable, especially when dealing with complex JSON datasets where I think this would be very useful.

For example, the code might end up looking like:
```
v := array_get(array_get(a.b.c, 1), 1) || 5
```
instead of
```
v := a.b.c[1][1] || 5
```

Quite a bit worse if you're dealing with something like:
```
v := array_get(array_get(array_get(a.b.c, 1), 1).b, 1).c || 5
```
instead of:
```
v := a[1][1].b[1].c || 5
```

d5:
Yeah, I know. I will close this PR but will create a new issue for this array index-get behavior so we can continue this discussion.

PR #85, opened 2019-02-06
add os.read_file()

ripienaar:


ripienaar:
used `readfile` not `read_file` since you already have `readlink`

d5:
It's not a strict rule, but, I tried to respect Go equivalent function's Camelcase. For example. `readlink` was `os.Readlink`. So I think it should be `read_file` as it's matching `ioutil.ReadFile`.

docs/stdlib-os.md
```diff
@@ -63,6 +63,7 @@ os := import("os")
 - `lookup_env(key string) => string/false`: retrieves the value of the environment variable named by the key.
 - `mkdir(name string, perm int) => error `: creates a new directory with the specified name and permission bits (before umask).
 - `mkdir_all(name string, perm int) => error `: creates a directory named path, along with any necessary parents, and returns nil, or else returns an error.
+- `readfile(name string) => array(byte)/error `: reads the contents of a file into a byte array
```
d5: should be `=> [byte]/error` to make it consistent with other docs.

d5:
It's not a strict rule, but, I tried to respect Go equivalent function's Camelcase. For example. `readlink` was `os.Readlink`. So I think it should be `read_file` as it's matching `ioutil.ReadFile`.

ripienaar:
update

ripienaar:
travis needs to be reran was an environmental failure

PR #103, opened 2019-02-19
Some improvements on VM error messages

### LINKED ISSUE #93: Line Number Info when compiled.Run() has error
Would be nice if the error returned by c.Run() would at least give the line number of the script where the error occurred.  A stack-trace would even be nicer in case there was some recursion, or user defined function call.

```	
if c, err := s.Compile(); err != nil {
	t.Errorf("compile error=%s",err )
} else {
	if err := c.Run(); err != nil {
		t.Errorf("error %s",err)
	}
}
```

d5:
- [x] type names in VM error messages
- [x] improve builtin function error messages
- [x] improve stdlib error messages
- [x] add `filename:line:column` to VM error messages
- [x] add `filename:line:column` to compiler error messages 
  - Example error message: `module1:43:10: export not allowed inside function`
  - Resolves #93. 
- [x] fix an issue with stdlib modules where they were not properly loaded when the code is executed from compiled binary


runtime/vm_test.go
```diff
@@ -104,16 +123,7 @@ func runVM(t *testing.T, file *ast.File, expected interface{}, symbols map[strin
...
-	_, trace, err := traceCompileRun(file, symbols, userModules)
-
-	defer func() {
-		if !ok {
-			t.Log("\n" + strings.Join(trace, "\n"))
-		}
-	}()
-
-	ok = assert.Error(t, err)
+func runVMError(file *ast.File, symbols map[string]objects.Object, userModules map[string]string) (trace []string, err error) {
```
golangcibot: `runVMError` is unused (from `deadcode`)

compiler/stdlib/func_typedefs_test.go
```diff
@@ -13,482 +13,487 @@ import (
...
 	assert.Equal(t, &objects.Error{Value: &objects.String{Value: "some error"}}, ret)
-	_, err = uf.Call()
+	_, err = funcCall(uf)
 	assert.Equal(t, objects.ErrWrongNumArguments, err)
 }
 
 func TestFuncASSRE(t *testing.T) {
 	uf := stdlib.FuncASSRE(func(a, b string) error { return nil })
-	ret, err := uf.Call(&objects.String{Value: "foo"}, &objects.String{Value: "bar"})
+	ret, err := funcCall(uf, &objects.String{Value: "foo"}, &objects.String{Value: "bar"})
```
golangcibot: ineffectual assignment to `ret` (from `ineffassign`)

earncef:
Error messages need at least a minimal trace. I've created a PR #105 with a simple `file:line:pos` added to the error.

Issue #106, opened 2019-02-20
Redundant OpExport?


earncef:
There's an OpExport defined in compiler/opcodes.go. Is this redundant?

d5:
Haha yup. That's from before I refactor module import.

PR #108, opened 2019-02-21
Stack trace for errors

earncef:
PR for stack trace.

 #93

### DIFF
--- a/runtime/vm.go
+++ b/runtime/vm.go
@@ -1373,11 +1373,6 @@ func (v *VM) Globals() []*objects.Object {
 	return v.globals
 }
 
-// FrameInfo returns the current function call frame information.
-func (v *VM) FrameInfo() (frameIndex, ip int) {
-	return v.framesIndex - 1, v.ip
-}
-
 func indexAssign(dst, src *objects.Object, selectors []*objects.Object) error {
 	numSel := len(selectors)
 
--- a/runtime/vm.go
+++ b/runtime/vm.go
@@ -84,7 +84,7 @@ func (v *VM) Abort() {
 }
 
 // Run starts the execution.
-func (v *VM) Run() error {
+func (v *VM) Run() (err error) {
 	// reset VM states
 	v.sp = 0
 	v.curFrame = &(v.frames[0])
@@ -104,7 +104,8 @@ mainloop:
 			v.ip += 2
 
 			if v.sp >= StackSize {
-				return ErrStackOverflow
+				err = ErrStackOverflow
+				break mainloop
 			}
 
 			v.stack[v.sp] = &v.constants[cidx]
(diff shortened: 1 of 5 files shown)


earncef:
I'm working on further optimization (unrelated to this). I will make a separate PR when I'm done.

earncef:
What do you think about introducing another Operator (something like OpBinaryOp) that'd be emitted before all the operators that call Object.BinaryOp()? 

I've been testing it out and it gets rid of a lot of redundant duplicate code without any loss of performance (a slight performance gain in fact, but probably not one to make it worth it just for the performance, especially for fib(35) which relies significantly on binary operations).

earncef:
Removed OpExport
#106

runtime/vm.go
```diff
@@ -1262,6 +1349,19 @@ mainloop:
...
+	if err != nil {
+		for v.ip < v.curIPLimit && v.framesIndex > 1 {
+			v.framesIndex--
+			v.curFrame = &v.frames[v.framesIndex-1]
+			v.curInsts = v.curFrame.fn.Instructions
+			v.curIPLimit = len(v.curInsts) - 1
+
+			filePos := v.fileSet.Position(v.curFrame.fn.SourceMap[v.curFrame.ip-1])
+			err = fmt.Errorf("%s\nin %s", err.Error(), filePos)
+		}
```
d5: I think this can be further reduced?
```golang
for v.framesIndex > 1 {
	v.framesIndex--
	v.curFrame = &v.frames[v.framesIndex-1]

	filePos := v.fileSet.Position(v.curFrame.fn.SourceMap[v.curFrame.ip-1])
	err = fmt.Errorf("%s\nin %s", err.Error(), filePos)
}
```

d5:
Tell me more about `OpBinaryOp`. I'm not sure if I understand what you're trying to do. Probably seeing some code is easier 😄 .

runtime/vm.go
d5: Actually, now that I think about this, we're mutating the VM state partially, so we probably need to either
1. remove v.FrameInfo() because it may return partially correct info, or, 
2. mutate everything as if VM actually returned from all function calls

The main purpose of v.FrameInfo was to see where the error occurred during test. But it's not needed anymore because we have stack trace in error.

earncef: Updated. 

True, I don't think v.FrameInfo adds any more useful information for debugging. Shall I remove it?

d5: Yes please

earncef:
I'll make a separate PR with the OpBinaryOp tomorrow.

runtime/vm.go
earncef: Done

PR #110, opened 2019-02-22
OpBinaryOp to remove redundant code in the VM

earncef:
This is what I was talking about yesterday. The `OpBinaryOp` is just a placeholder that removes the necessity of all the redundant code for each operation that calls Object.BinaryOp().

While this is possible to do as `case compiler.OpAdd, compiler.OpSub, compiler.OpMul, ...`, that ends up reducing performance of `fib(35)` recursive by about 1-2% whereas the `OpBinaryOp` version is slightly (negligibly) faster than the duplicated code.

### DIFF
--- a/compiler/compiler.go
+++ b/compiler/compiler.go
@@ -123,6 +123,7 @@ func (c *Compiler) Compile(node ast.Node) error {
 				return err
 			}
 
+			c.emit(node, OpBinaryOp)
 			c.emit(node, OpGreaterThan)
 
 			return nil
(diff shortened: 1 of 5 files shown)


d5:
I want to think through this a bit, but, using an Opcode as a parameter for another Opcode doesn't feel right in many ways. They should really be Opcode. But, again I will think more about this. Thanks for the proposal!!

earncef:
I know! That’s why I added this as a separate PR. It’s not essential either but at the same time it gets rid of a large amount of redundant code.

earncef:
#132  is a better solution.

PR #113, opened 2019-02-25
Fixed panic when bailout is triggered in parser.ParseFile()

earncef:
Bailout shouldn't cause panic but return errors as in parser/parse_file.go


earncef:
I've also removed the redundant "parse error" text in script.Compile() for parse errors. This prefix is not already included for all parse errors.

d5:
Good catch. I don't know if you noticed or not, but, originally, the intention was to catch it from [parser.ParseFile](https://github.com/d5/tengo/blob/master/compiler/parser/parse_file.go#L16), and, force using it. But apparently that's the case anymore haha.

Issue #114, opened 2019-02-25
String keys not allowed in map definition


earncef:
The following format is not allowed:
```
m := {
    "a": 1
}
```

This is not consistent with Go and can also cause unnecessary clutter in the code when the key is not alphanumeric. For instance:
```
m := {
    "@a": 1
}
```

earncef:
I'll have a PR for this soon.

PR #118, opened 2019-02-26
String keys allowed for map definitions.

earncef:
#114 

One comment:
p.expects allows only one token and I'm not sure whether token.Ident would be the best option, but I have left it as is for now.

### DIFF
--- a/compiler/parser/parser.go
+++ b/compiler/parser/parser.go
@@ -1016,15 +1016,27 @@ func (p *Parser) parseMapElementLit() *ast.MapElementLit {
 	}
 
 	// key: read identifier token but it's not actually an identifier
-	ident := p.parseIdent()
+	pos := p.pos
+	name := "_"
+
+	if p.token == token.Ident {
+		name = p.tokenLit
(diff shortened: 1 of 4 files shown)


compiler/parser/parser.go
```diff
@@ -1016,15 +1016,27 @@ func (p *Parser) parseMapElementLit() *ast.MapElementLit {
...
+
+	if p.token == token.Ident {
+		name = p.tokenLit
+		p.next()
+	} else if p.token == token.String {
+		v, _ := strconv.Unquote(p.tokenLit)
+		name = v
+		p.next()
+	} else {
+		p.expect(token.Ident)
```
d5: What about this?
```golang
p.errorExpected(pos, "map key") // or maybe "string"?
p.next()
```

Then, you might want to move `p.next()` to below `if-else if-else` statement.

earncef: Yeah, that looks better. "map key" is perhaps more suitable then "string" as both token.String and token.Ident are allowed.

PR #124, opened 2019-03-01
script.Add accepts a CallableFunc value

### LINKED ISSUE #77: script.Add should accept a CallableFunc value
To be able to directly define a user function
`s.Add("myFunc", myFunc)`

Now, you need to do
`s.Add("myFunc", &objects.UserFunction{Value: myFunc})`

earncef:
fixes #77 

I've the conversion in scripts.Add instead of objects.FromInterface as it is not clear that objects.FromInterface will never be called for builtins.

### DIFF
--- a/objects/conversion.go
+++ b/objects/conversion.go
@@ -243,6 +243,10 @@ func FromInterface(v interface{}) (Object, error) {
 		return &Time{Value: v}, nil
 	case Object:
 		return v, nil
+	case CallableFunc:
+		return &UserFunction{Value: v}, nil
+	case func(...Object) (Object, error):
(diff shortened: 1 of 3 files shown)


d5:
I think it's safe to assume that builtin functions will be defined explicitly, not through `objects.FromInterface`. We can move it inside `objects.FromInterface`.

d5:
Btw, not sure if you know this or not, but, you can auto-close related issue by using 'fixes' or 'resolves': https://help.github.com/en/articles/closing-issues-using-keywords

earncef:
Done

d5:
Thanks @earncef !

PR #126, opened 2019-03-02
Decoupled stdlib from vm, script and compiler

earncef:
stdlib is fully optional. this is in prep for #68

### DIFF
--- a/cmd/tengomin/main.go
+++ b/cmd/tengomin/main.go
@@ -189,7 +189,7 @@ func runREPL(in io.Reader, out io.Writer) {
 	var constants []objects.Object
 
 	for {
-		_, _ = fmt.Fprintf(out, replPrompt)
+		_, _ = fmt.Fprint(out, replPrompt)
 
 		scanned := stdin.Scan()
 		if !scanned {
--- a/cmd/tengomin/main.go
+++ b/cmd/tengomin/main.go
@@ -17,7 +17,6 @@ import (
 	"github.com/d5/tengo/compiler/source"
 	"github.com/d5/tengo/objects"
 	"github.com/d5/tengo/runtime"
-	"github.com/d5/tengo/stdlib"
 )
 
 const (
@@ -26,12 +25,10 @@ const (
 )
 
 var (
-	compileOutput  string
-	showHelp       bool
-	showVersion    bool
-	version        = "dev"
-	bm             map[string]bool
-	builtinModules map[string]*objects.Object
+	compileOutput string
+	showHelp      bool
+	showVersion   bool
+	version       = "dev"
 )
 
 func init() {
@@ -48,13 +45,6 @@ func main() {
 	} else if showVersion {
 		fmt.Println(version)
 		return
-	}
-
-	bm = make(map[string]bool, len(stdlib.Modules))
-	builtinModules = make(map[string]*objects.Object, len(stdlib.Modules))
-	for k, mod := range stdlib.Modules {
-		bm[k] = true
-		builtinModules[k] = objectPtr(mod)
 	}
 
 	inputFile := flag.Arg(0)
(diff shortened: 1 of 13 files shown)


runtime/vm_test.go
```diff
@@ -31,30 +31,46 @@ func expect(t *testing.T, input string, expected interface{}) {
...
-	runVMError(t, input, nil, nil, expected)
+	runVMError(t, input, nil, nil, nil, expected)
 }
 
 func expectErrorWithUserModules(t *testing.T, input string, userModules map[string]string, expected string) {
-	runVMError(t, input, nil, userModules, expected)
+	runVMError(t, input, nil, userModules, nil, expected)
+}
+
+func expectErrorWithBuiltinModules(t *testing.T, input string, builtinModules map[string]*objects.Object, expected string) {
```
golangcibot: `expectErrorWithBuiltinModules` is unused (from `deadcode`)

runtime/vm_test.go
```diff
@@ -31,30 +31,46 @@ func expect(t *testing.T, input string, expected interface{}) {
...
+	runVMError(t, input, nil, nil, builtinModules, expected)
+}
+
+func expectErrorWithUserAndBuiltinModules(t *testing.T, input string, userModules map[string]string, builtinModules map[string]*objects.Object, expected string) {
```
golangcibot: `expectErrorWithUserAndBuiltinModules` is unused (from `deadcode`)

runtime/vm_test.go
```diff
@@ -171,8 +187,15 @@ func traceCompileRun(file *ast.File, symbols map[string]objects.Object, userModu
 		symTable.DefineBuiltin(idx, fn.Name)
 	}
 
+	bm := make(map[string]bool)
+	if builtinModules != nil {
```
golangcibot: S1031: unnecessary nil check around range (from `gosimple`)

runtime/vm_test.go
```diff
@@ -171,8 +187,15 @@ func traceCompileRun(file *ast.File, symbols map[string]objects.Object, userModu
...
+		for k, _ := range builtinModules {
```
golangcibot: S1005: should omit value from range; this loop is equivalent to `for k := range ...` (from `gosimple`)

d5:
Hey, nice work as always. I have 2 comments:
1. I think we don't have to remove `stdlib` from tests. Test packages will not be included when Tengo is imported by other packages. 
2. What's your plan for `cmd/tengo`? We will have to ship 2 different binaries? One with stdlib, another without stdlib?

earncef:
1. I removed it from tests and moved some to stdlib (tests for os) not because of importing but for improved separation of concerns. The tests in vm should ideally be testing the vm itself and not the stdlib and the same goes for scripts. Stdlib now has some of the tests that were originally in the vm that test the stdlib functionality. I think stdlib might even do with some more tests but that’s a separate issue altogether.
2. Yes, IMO binaries would probably be the most straightforward way to deal with this.

earncef:
Also, I’ll try to work on improving the tests in the packages at some point. There’s some redundant tests that I think can be cleaned up and the focus narrowed a bit more tightly.

d5:
Ok. Makes sense. We will have to update documentations, `cmd/tengo` _(likely just another package to include all stdlib?)_, and, also the build scripts accordingly.

d5:
@earncef hey what's your email address? I'd like to invite you to Slack group.

earncef:
earncefs-at-yahoo

cmd/tengomin/main.go
```diff
@@ -0,0 +1,293 @@
...
+
+	symbolTable := compiler.NewSymbolTable()
+	for idx, fn := range objects.Builtins {
+		symbolTable.DefineBuiltin(idx, fn.Name)
+	}
+
+	var constants []objects.Object
+
+	for {
+		_, _ = fmt.Fprintf(out, replPrompt)
```
golangcibot: SA1006: printf-style function with dynamic format string and no further arguments should use print-style function instead (from `staticcheck`)

earncef:
Tengo binary is quite small. I'm not entirely sure the second binary without stdlib bundled is worth it. A flag in the main binary to disable stdlib might be enough.

PR #134, opened 2019-03-07
concurrent execution of script.Compiled

### LINKED ISSUE #133: Is it possible to reuse the pre-compiled VM in concurrent environment
My use-case is to have an external script define the execution logic (data mutation/enhancements) which can happen on concurrent requests... (think about an HTTP server with Middleware mutating request headers, where the logic for mutation is defined by a custom external script).  The server supports multiple requests coming in at once, and for performance reasons I would like to avoid having to instantiate and compile a new VM instance for every request.

Is it possible to have one VM instance be compiled and re-used in a safe concurrent calls environment, or do I have to create new VM per request, as there is possible race condition with `c.Set()` and `c.Get()` mutating the same global variable in the script from concurrent executions.

d5:
Use `Compiled.Clone()` to create a copy of Compiled instance. Cloned copies are safe for concurrent use by multiple goroutines.

```golang
// compile the script once
compiled, _ := myscript.Compile()

// execute from multiple goroutines
for i := 0; i < 100; i++ {
    go func(compiled *script.Compiled) {
        // inputs
        _ = compiled.Set("a", rand.Intn(10))
        _ = compiled.Set("b", rand.Intn(10))
        _ = compiled.Set("c", rand.Intn(10))
        
        if err := compiled.Run(); err != nil {
            panic(err)
        }
        
        // outputs
        d = compiled.Get("d").Int()
        e = compiled.Get("e").Int()
    }(compiled.Clone()) // Pass the cloned copy of Compiled
}
```

- [x] add `Compiled.Clone` function
- [x] update documentation

This resolves #133.

### DIFF
--- a/script/compiled.go
+++ b/script/compiled.go
@@ -3,6 +3,7 @@ package script
 import (
 	"context"
 	"fmt"
+	"sync"
 
 	"github.com/d5/tengo/compiler"
 	"github.com/d5/tengo/objects"
@@ -12,26 +13,41 @@ import (
 // Compiled is a compiled instance of the user script.
 // Use Script.Compile() to create Compiled object.
 type Compiled struct {
-	symbolTable *compiler.SymbolTable
-	machine     *runtime.VM
+	globalIndexes    map[string]int // global symbol name to index
+	bytecode         *compiler.Bytecode
+	globals          []*objects.Object
+	builtinFunctions []objects.Object
+	builtinModules   map[string]*objects.Object
+	maxAllocs        int64
+	lock             sync.RWMutex
 }
 
 // Run executes the compiled script in the virtual machine.
 func (c *Compiled) Run() error {
-	return c.machine.Run()
+	c.lock.Lock()
+	defer c.lock.Unlock()
+
+	v := runtime.NewVM(c.bytecode, c.globals, c.builtinFunctions, c.builtinModules, c.maxAllocs)
+
+	return v.Run()
 }
 
 // RunContext is like Run but includes a context.
 func (c *Compiled) RunContext(ctx context.Context) (err error) {
+	c.lock.Lock()
+	defer c.lock.Unlock()
+
+	v := runtime.NewVM(c.bytecode, c.globals, c.builtinFunctions, c.builtinModules, c.maxAllocs)
+
 	ch := make(chan error, 1)
 
 	go func() {
-		ch <- c.machine.Run()
+		ch <- v.Run()
 	}()
 
 	select {
 	case <-ctx.Done():
-		c.machine.Abort()
+		v.Abort()
 		<-ch
 		err = ctx.Err()
 	case err = <-ch:
@@ -40,14 +56,42 @@ func (c *Compiled) RunContext(ctx context.Context) (err error) {
 	return
 }
 
+// Clone creates a new copy of Compiled.
+// Cloned copies are safe for concurrent use by multiple goroutines.
+func (c *Compiled) Clone() *Compiled {
+	c.lock.Lock()
+	defer c.lock.Unlock()
+
+	clone := &Compiled{
+		globalIndexes:    c.globalIndexes,
+		bytecode:         c.bytecode,
+		globals:          make([]*objects.Object, len(c.globals)),
+		builtinFunctions: c.builtinFunctions,
+		builtinModules:   c.builtinModules,
+		maxAllocs:        c.maxAllocs,
+	}
+
+	// copy global objects
+	for idx, g := range c.globals {
+		if g != nil {
+			clone.globals[idx] = objectPtr((*g).Copy())
+		}
+	}
+
+	return clone
+}
+
 // IsDefined returns true if the variable name is defined (has value) before or after the execution.
 func (c *Compiled) IsDefined(name string) bool {
-	symbol, _, ok := c.symbolTable.Resolve(name)
+	c.lock.RLock()
+	defer c.lock.RUnlock()
+
+	idx, ok := c.globalIndexes[name]
 	if !ok {
 		return false
 	}
 
-	v := c.machine.Globals()[symbol.Index]
+	v := c.globals[idx]
 	if v == nil {
 		return false
 	}
@@ -57,11 +101,13 @@ func (c *Compiled) IsDefined(name string) bool {
 
 // Get returns a variable identified by the name.
 func (c *Compiled) Get(name string) *Variable {
+	c.lock.RLock()
+	defer c.lock.RUnlock()
+
 	value := &objects.UndefinedValue
 
-	symbol, _, ok := c.symbolTable.Resolve(name)
-	if ok && symbol.Scope == compiler.ScopeGlobal {
-		value = c.machine.Globals()[symbol.Index]
+	if idx, ok := c.globalIndexes[name]; ok {
+		value = c.globals[idx]
 		if value == nil {
 			value = &objects.UndefinedValue
 		}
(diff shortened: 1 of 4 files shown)


script/compiled.go
```diff
@@ -40,14 +56,42 @@ func (c *Compiled) RunContext(ctx context.Context) (err error) {
...
+// Cloned copies are safe for concurrent use by multiple goroutines.
+func (c *Compiled) Clone() *Compiled {
+	c.lock.Lock()
+	defer c.lock.Unlock()
+
+	clone := &Compiled{
+		globalIndexes:    c.globalIndexes,
+		bytecode:         c.bytecode,
+		globals:          make([]*objects.Object, len(c.globals)),
+		builtinFunctions: c.builtinFunctions,
```
aantono: Is there a need to also copy/clone the `builtinFunctions` and `builtinModules`, or are they "thread-safe"?

d5: All states are read-only by VM, except for globals.

aantono: What if they are being mutated by the calling host?

I'm thinking about a usecase where one makes a `compiled.Clone()` and then modifies the functions (or something to that extent).

aantono: Ah, looks like one can't mutate them from outside `Compiled`, only when compiling the `Script`, so once you get a `Compiled` instance, it's all final at that point... ;)

d5: Yeah, I know that's possible technically. But, in a way, all Compiled instances are considered to be outcome of the same Script, which had defined all globals, builtins, and, etc. So, conceptually, those states are defined/managed by Script, not by Compiled. And, if one wants to change those states, it's encouraged to create a new Script instance.

aantono: Makes sense... as long as it's ok to mutate the global variables, the functions, etc... should stay static.  The best way, to be honest, would be to enable support for defining functions in the script with arguments, and be able to invoke those functions from the host by passing the function arguments externally... 😉

d5: Can you elaborate?

In most of my use cases, I actually force the user script to `export` a function and load it as a module from a wrapper code. Look at this example: https://pbr.cloud/simulator _(remember this is one of work-in-progress projects)_.

aantono: My example would be something like this. 
Imagine you have a user/tengo script loaded from disk that has a function defined, something simple like:
```
Handle := func(req, resp) {
    return {req: req, resp: resp}
}
```
In the host, we load the script and then try to invoke the `Handle(Request, Response)` function, but by passing the `Request` and `Response` objects from the host:
```
func main() {
    req := "Imagine an HTTP Request struct"
    resp := "Imagine an HTTP Response struct"
    s := script.New([]byte(code))
    c, _ := s.Compile()
    if _, err := c.RunWithArgs("Handle", req, resp); err != nil {
        panic(err)
    }
}
```

d5: Yeah, that's a frequent usage pattern. Calling a function (defined in global scope) directly from the host can be useful _(for another PR)_. But, regardlessly, the arguments need to be simple values (that can be automatically converted to existing primitive types), or, it needs a wrapper that implements `objects.Object` interface.

Here's an example HTTP request handling context object that I use in another project:

```golang
package server

import (
	"bytes"
	"io/ioutil"
	"net/http"

	"github.com/d5/tengo/compiler/token"
	"github.com/d5/tengo/objects"
)

type ExecContext struct {
	reqHeader   *objects.ImmutableMap
	queryParams *objects.ImmutableMap
	reqBody     *objects.Bytes
	status      int64
	resHeader   map[string]string
	resBody     bytes.Buffer
}

func NewExecContext(req *http.Request) (*ExecContext, error) {
	// request header
	reqHeader := make(map[string]objects.Object)
	for key, values := range req.Header {
		if len(values) > 0 {
			reqHeader[key] = &objects.String{Value: values[0]}
		}
	}

	// request query params
	queryParams := make(map[string]objects.Object)
	for name, values := range req.URL.Query() {
		if name != "s" && len(values) > 0 {
			queryParams[name] = &objects.String{Value: values[0]}
		}
	}

	// request body
	reqBody, err := ioutil.ReadAll(req.Body)
	if err != nil {
		return nil, err
	}

	return &ExecContext{
		reqHeader:   &objects.ImmutableMap{Value: reqHeader},
		queryParams: &objects.ImmutableMap{Value: queryParams},
		reqBody:     &objects.Bytes{Value: reqBody},
		status:      http.StatusOK,
		resHeader:   make(map[string]string),
	}, nil
}

func (c *ExecContext) TypeName() string {
	return "context"
}

func (c *ExecContext) String() string {
	return "<context>"
}

func (c *ExecContext) BinaryOp(op token.Token, rhs objects.Object) (objects.Object, error) {
	return nil, objects.ErrInvalidOperator
}

func (c *ExecContext) IsFalsy() bool {
	return false
}

func (c *ExecContext) Equals(another objects.Object) bool {
	return false
}

func (c *ExecContext) Copy() objects.Object {
	return c // always shared
}

func (c *ExecContext) IndexGet(index objects.Object) (objects.Object, error) {
	indexStr, ok := objects.ToString(index)
	if !ok {
		return nil, objects.ErrInvalidIndexType
	}

	switch indexStr {
	case "req_header":
		return c.reqHeader, nil
	case "req_body":
		return c.reqBody, nil
	case "query":
		return c.queryParams, nil
	case "set_status":
		return &objects.UserFunction{Value: c.setStatus}, nil
	case "set_header":
		return &objects.UserFunction{Value: c.setHeader}, nil
	case "delete_header":
		return &objects.UserFunction{Value: c.deleteHeader}, nil
	case "write":
		return &objects.UserFunction{Value: c.write}, nil
	default:
		return objects.UndefinedValue, nil
	}
}

func (c *ExecContext) Status() int {
	return int(c.status)
}

func (c *ExecContext) ResHeader() map[string]string {
	return c.resHeader
}

func (c *ExecContext) ResBody() []byte {
	return c.resBody.Bytes()
}

func (c *ExecContext) setStatus(args ...objects.Object) (ret objects.Object, err error) {
	if len(args) != 1 {
		return nil, objects.ErrWrongNumArguments
	}

	status, ok := objects.ToInt64(args[0])
	if !ok {
		return nil, objects.ErrInvalidArgumentType{
			Name:     "status",
			Expected: "int(compatible)",
			Found:    args[0].TypeName(),
		}
	}

	c.status = status

	return objects.UndefinedValue, nil
}

func (c *ExecContext) setHeader(args ...objects.Object) (ret objects.Object, err error) {
	if len(args) != 2 {
		return nil, objects.ErrWrongNumArguments
	}

	key, ok := objects.ToString(args[0])
	if !ok {
		return nil, objects.ErrInvalidArgumentType{
			Name:     "key",
			Expected: "string(compatible)",
			Found:    args[0].TypeName(),
		}
	}

	value, ok := objects.ToString(args[1])
	if !ok {
		return nil, objects.ErrInvalidArgumentType{
			Name:     "value",
			Expected: "string(compatible)",
			Found:    args[1].TypeName(),
		}
	}

	c.resHeader[key] = value

	return objects.UndefinedValue, nil
}

func (c *ExecContext) deleteHeader(args ...objects.Object) (ret objects.Object, err error) {
	if len(args) != 1 {
		return nil, objects.ErrWrongNumArguments
	}

	key, ok := objects.ToString(args[0])
	if !ok {
		return nil, objects.ErrInvalidArgumentType{
			Name:     "key",
			Expected: "string(compatible)",
			Found:    args[0].TypeName(),
		}
	}

	delete(c.resHeader, key)

	return objects.UndefinedValue, nil
}

func (c *ExecContext) write(args ...objects.Object) (ret objects.Object, err error) {
	if len(args) != 1 {
		return nil, objects.ErrWrongNumArguments
	}

	if v, ok := args[0].(*objects.Bytes); ok {
		_, err := c.resBody.Write(v.Value)
		if err != nil {
			return &objects.Error{Value: &objects.String{Value: err.Error()}}, nil
		}
	} else {
		v, ok := objects.ToString(args[0])
		if !ok {
			return nil, objects.ErrInvalidArgumentType{
				Name:     "value",
				Expected: "bytes/string(compatible)",
				Found:    args[0].TypeName(),
			}
		}
		_, err := c.resBody.WriteString(v)
		if err != nil {
			return &objects.Error{Value: &objects.String{Value: err.Error()}}, nil
		}
	}

	return objects.UndefinedValue, nil
}
```

Then, I create an instance and pass it to the script like this:

```golang
ctx := createContext(req, res) // *ExecContext which implements objects.Object
compiled.Set("ctx", ctx)
compiled.Run()

for key, value := range ctx.ResHeader() {
	res.Header().Set(key, value)
}
res.WriteHeader(ctx.Status())
_, _ = res.Write(ctx.ResBody())
```

Also the user code is loaded as a module by this code wrapper:

```golang
s := script.New([]byte(`import("script")(ctx)`)
```

And the actual user code should look like

```golang
rand := import("rand")

export func(ctx) {
    ctx.set_status(200)
    
    ctx.set_header("Content-Type", "text/plain")
    
    ctx.write(string(rand.intn(100)))
}
```

aantono: Makes sense... I guess it's just not as "elegant"...  Also forces you to do `compiled.Clone()` for concurrent executions, as the variable `ctx` gets bound to global scope, instead of the method stack, so makes things a bit complicated. 😉

PR #140, opened 2019-03-13
Removed double indirection for massive performance gains

earncef:
Double indirection is gone for all of the VM except free vars. For free vars, the object is stored in objects.FreeVar.Value and then used accordingly. 

OLD:
```
fibonacci(35)
Result:  9227465
Go:      55.639295ms
Parser:  50.693µs
Compile: 105.823µs
VM:      4.337952964s
fibonacci(35) (tail-call #1)
Result:  9227465
Go:      76.950689ms
Parser:  24.202µs
Compile: 71.432µs
VM:      4.44593454s
fibonacci(35) (tail-call #2)
Result:  9227465
Go:      509ns
Parser:  33.497µs
Compile: 88.749µs
VM:      18.793µs
```

NEW:
```
fibonacci(35)
Result:  9227465
Go:      57.868119ms
Parser:  48.051µs
Compile: 103.497µs
VM:      3.155774638s
fibonacci(35) (tail-call #1)
Result:  9227465
Go:      77.439483ms
Parser:  23.099µs
Compile: 32.781µs
VM:      3.256651909s
fibonacci(35) (tail-call #2)
Result:  9227465
Go:      351ns
Parser:  19.647µs
Compile: 26.19µs
VM:      19.461µs
```


runtime/vm.go
```diff
@@ -1209,18 +1289,29 @@ func indexAssign(dst, src *objects.Object, selectors []*objects.Object) error {
 	return nil
 }
 
+func (v *VM) printStack(sp int) {
```
golangcibot: U1000: func `(*VM).printStack` is unused (from `unused`)

compiler/opcodes.go
```diff
@@ -48,8 +48,11 @@ const (
 	OpSetLocal                       // Set local variable
 	OpDefineLocal                    // Define local variable
 	OpSetSelLocal                    // Set local variable using selectors
+	OpGetFreePtr                     // Get free variable pointer object
 	OpGetFree                        // Get free variables
 	OpSetFree                        // Set free variables
+	OpGetLocalPtr                    // Get local variable as a pointer
+	OpSetLocalPtr                    // Set local variable as a pointer
```
d5: I noticed that `OpSetLocalPtr` and `OpGetLocalPtr` can be combined into one opcode as they're always used together. But, when I tested with combined opcode, for some reasons, the benchmark was **slower** (adding 0.2-0.4 seconds) even though fib(35) does not even use them. Combined code looks like this:
```
			localIndex := int(v.curInsts[v.ip+1])
			v.ip++

			if v.curFrame.localFreeVars == nil {
				v.curFrame.localFreeVars = make(map[int]*objects.FreeVar)
			}

			sp := v.curFrame.basePointer + localIndex

			val := v.stack[sp]

			if obj, ok := val.(*objects.FreeVar); ok {
				v.curFrame.localFreeVars[localIndex] = obj
			} else {
				v.curFrame.localFreeVars[localIndex] = &objects.FreeVar{Value: &val}
			}

			v.stack[sp] = v.curFrame.localFreeVars[localIndex]

			if v.sp >= StackSize {
				v.err = ErrStackOverflow
				return
			}

			v.stack[v.sp] = v.curFrame.localFreeVars[localIndex]
			v.sp++
```
Definitely something to investigate further.

objects/free_var.go
```diff
@@ -0,0 +1,41 @@
+package objects
+
+import (
+	"github.com/d5/tengo/compiler/token"
+)
+
+// FreeVar represents a free variable.
+type FreeVar struct {
```
d5: What about `ObjectPtr` instead of `FreeVar`?

runtime/vm.go
```diff
@@ -22,15 +22,15 @@ const (
...
+	undefinedPtr = objects.UndefinedValue
 )
 
 // VM is a virtual machine that executes the bytecode compiled by Compiler.
 type VM struct {
 	constants      []objects.Object
-	stack          []*objects.Object
+	stack          []objects.Object
 	sp             int
 	globals        []*objects.Object
```
d5: this can be `[]objects.Object` now.

d5: also `vm.builtinModules` and `script.Variable.value` don't have to be object pointers anymore.

runtime/vm.go
```diff
@@ -207,7 +207,7 @@ func (v *VM) run() {
 				return
 			}
 
-			if (*left).Equals(*right) {
+			if (left).Equals(right) {
```
d5: parenthesis are redundant

runtime/vm.go
```diff
@@ -379,19 +379,25 @@ func (v *VM) run() {
...
+			for i := 0; i < numSelectors; i++ {
+				obj := v.stack[v.sp-numSelectors+i]
+				selectors[i] = &obj
+			}
+
 			val := v.stack[v.sp-numSelectors-1]
 			v.sp -= numSelectors + 1
 
-			if e := indexAssign(v.globals[globalIndex], val, selectors); e != nil {
+			if e := indexAssign(v.globals[globalIndex], &val, selectors); e != nil {
```
d5: `indexAssign()` can take non-pointers now:
```
func indexAssign(dst, src objects.Object, selectors []objects.Object) error {
```

runtime/vm.go
d5: I see more redundant parenthesis.

runtime/vm.go
d5: if `globals` are non-pointer slice, there are some code that be can be reduced. For example, the following code in `OpSetGlobal` case:

```golang
obj := v.stack[v.sp]
v.globals[globalIndex] = obj
```

can be

```golang
v.globals[globalIndex] = v.stack[v.sp]
```

d5: plus, we can remove `objectPtr()` utility functions in several different places.

earncef:
I'll push these changes tomorrow.

d5:
I think we can combine `OpGetLocalPtr` and `OpSetLocalPtr` like this:

```golang
case compiler.OpGetLocalPtr:
	localIndex := int(v.curInsts[v.ip+1])
	v.ip++

	sp := v.curFrame.basePointer + localIndex
	val := v.stack[sp]

	var freeVar *objects.FreeVar
	if obj, ok := val.(*objects.FreeVar); ok {
		freeVar = obj
	} else {
		freeVar = &objects.FreeVar{Value: &val}
		v.stack[sp] = freeVar
	}

	if v.sp >= StackSize {
		v.err = ErrStackOverflow
		return
	}

	v.stack[v.sp] = freeVar
	v.sp++
```

Also, if we do this, we can remove `Frame.localFreeVars` as well.

Why this slows down `fib35()` is still a mystery though.

script/script.go
```diff
@@ -216,7 +216,7 @@ func (s *Script) prepCompile() (symbolTable *compiler.SymbolTable, builtinModule
 		builtinModules[name] = true
 	}
 
-	globals = make([]*objects.Object, runtime.GlobalsSize, runtime.GlobalsSize)
+	globals = make([]objects.Object, runtime.GlobalsSize, runtime.GlobalsSize)
```
golangcibot: S1019: should use make([]objects.Object, runtime.GlobalsSize) instead (from `gosimple`)

runtime/vm.go
earncef: Removed all I could find.

earncef:
All done

PR #144, opened 2019-03-15
Added println

### LINKED ISSUE #142: add println builtin function
and `print` should not append newline character at the end

earncef:
fixes #142

### DIFF
--- a/objects/builtin_print.go
+++ b/objects/builtin_print.go
@@ -10,6 +10,19 @@ import (
 func builtinPrint(args ...Object) (Object, error) {
 	for _, arg := range args {
 		if str, ok := arg.(*String); ok {
+			fmt.Print(str.Value)
+		} else {
+			fmt.Print(arg.String())
+		}
+	}
+
+	return nil, nil
+}
+
+// print(args...)
+func builtinPrintln(args ...Object) (Object, error) {
+	for _, arg := range args {
+		if str, ok := arg.(*String); ok {
 			fmt.Println(str.Value)
 		} else {
 			fmt.Println(arg.String())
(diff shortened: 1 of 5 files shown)


d5:
Now that we're adding `println`, I think we should try to match `fmt.Print()` and `fmt.Println()`. So, it could look like this;

```golang
	var printArgs []interface{}
	l := 0
	for _, arg := range args {
		s, _ := objects.ToString(arg)
		slen := len(s)
		if l+slen > tengo.MaxStringLen { // make sure length does not exceed the limit
			return nil, objects.ErrStringLimit
		}
		l += slen

		printArgs = append(printArgs, s)
	}

	_, _ = fmt.Print(printArgs...) // or fmt.Println()
```

Regarding the size limit, it seems that `fmt.Print` and `fmt.Println` have different rules of adding spaces between operands, which means the actual string size may be larger than `l`. But I'm not sure if we need to enforce to that level.

earncef:
Added the size limit. I don't think accounting for spaces in the size is necessary.

objects/builtin_print.go
```diff
@@ -6,16 +6,46 @@ import (
...
+		printArgs = append(printArgs, s)
+	}
+
+	return printArgs, nil
+}
+
+// print(args...)
+func builtinPrint(args ...Object) (Object, error) {
+	printArgs, err := getPrintArgs(args...)
+
```
d5: please remove empty line here and another one in Println

d5:
Also, can you update builtin documentation too?

earncef:
I have already updated documentation

PR #145, opened 2019-03-15
Refactored duplicate REPL code

earncef:


repl/repl.go
```diff
@@ -0,0 +1,296 @@
...
+
+	symbolTable := compiler.NewSymbolTable()
+	for idx, fn := range objects.Builtins {
+		symbolTable.DefineBuiltin(idx, fn.Name)
+	}
+
+	var constants []objects.Object
+
+	for {
+		_, _ = fmt.Fprintf(out, replPrompt)
```
golangcibot: SA1006: printf-style function with dynamic format string and no further arguments should use print-style function instead (from `staticcheck`)

cmd/tengo/main.go
```diff
@@ -1,23 +1,11 @@
...
-	"github.com/d5/tengo/compiler/parser"
-	"github.com/d5/tengo/compiler/source"
-	"github.com/d5/tengo/objects"
-	"github.com/d5/tengo/runtime"
+	"github.com/d5/tengo/repl"
 	"github.com/d5/tengo/stdlib"
+	"github.com/d5/tengo/objects"
 )
 
 const (
```
d5: `sourceFileExt` and `replPrompt not needed.

d5: `bm` also not needed

repl/repl.go
```diff
@@ -0,0 +1,312 @@
+package repl
```
d5: Can we change this to `cli` instead? REPL is just one of the functions that tengo CLI does. `cli/cli.go`

repl/repl.go
```diff
@@ -0,0 +1,312 @@
...
+	InputFile      string
+
+	//Builtin modules
+	BuiltinModules map[string]objects.Object
+}
+
+var (
+	bm             map[string]bool
+	builtinModules map[string]objects.Object
+	version        = "dev"
```
d5: `version` should be passed by main package as the build (goreleaser) set the version value of main package (`cmd/tengo` and `cmd/tengomin`).

PR #151, opened 2019-03-18
update README

d5:
- remove Patreon badge
- update example code and link
- add some use cases


README.md
```diff
@@ -25,11 +24,11 @@ sum := func(init, seq) {
     return init
 }
 
-n := sum(0, [1, 2, 3])   // == 6
-s := sum("", [1, 2, 3])  // == "123"
+println(sum(0, [1, 2, 3]))   // "6"
+println(sum("", [1, 2, 3]))  // "123"
```
earncef: fmt.println

d5: The playground runtime has its own `println` builtin function (that writes to internal buffer instead of stdout), and, "fmt" module is disabled. You can try it out in the link: https://tengolang.com/?s=3243fd27cd518aef287d16d59349e15a474fd76c

earncef: Then perhaps a comment explaining that the code would only work on playground? I can imagine someone installing tengo binary and following the readme to try println and failing.

d5:
Updated the tengolang.com editor to use modified 'fmt' module. So the sample code can be easily copy-pasted.

PR #152, opened 2019-03-19
fix bytecode encoding/decoding

d5:
- [x] register all possible object types
- [x] fix/normalize Bool/Undefined object values

### DIFF
--- a/compiler/bytecode_optimize.go
+++ b/compiler/bytecode_optimize.go
@@ -25,7 +25,7 @@ func (b *Bytecode) RemoveDuplicates() {
 			indexMap[curIdx] = len(deduped)
 			deduped = append(deduped, c)
 		case *objects.ImmutableMap:
-			modName := c.Value["__module_name__"].(*objects.String).Value
+			modName := moduleName(c)
 			newIdx, ok := immutableMaps[modName]
 			if modName != "" && ok {
 				indexMap[curIdx] = newIdx
@@ -72,7 +72,7 @@ func (b *Bytecode) RemoveDuplicates() {
 				deduped = append(deduped, c)
 			}
 		default:
-			panic(fmt.Errorf("invalid constant type: %s", c.TypeName()))
+			panic(fmt.Errorf("unsupported top-level constant type: %s", c.TypeName()))
 		}
 	}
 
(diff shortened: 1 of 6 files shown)


compiler/bytecode_optimize.go
```diff
@@ -25,7 +25,7 @@ func (b *Bytecode) RemoveDuplicates() {
 			indexMap[curIdx] = len(deduped)
 			deduped = append(deduped, c)
 		case *objects.ImmutableMap:
-			modName := c.Value["__module_name__"].(*objects.String).Value
+			modName := moduleName(c)
 			newIdx, ok := immutableMaps[modName]
```
earncef: It's actually possible to include a module without a name:
```
s.SetImports(map[string]objects.Importable{
    "": stdlib.BuiltinModules["math"],
})
```

which then allows
```
import("")
```

In addition, I think it might be better to change the parameter for `script.SetImports()` and `importModules` parameter in `compiler.NewCompiler()` from `map[string]objects.Importable` to `[]objects.Importable` to avoid unexpected scenarios such as accidental import of `math` as `fmt` or something of the like.

d5: Good point that an empty module name is allowed in the current implementation. I think we should disallow them at the compiler level (or even parser level).

To your second point, can you give me an example of accidental imports? The reasons I made it a map were
1. a module can be imported using any arbitrary names (e.g. `math := import("fmt")` is not a good example but maybe one wants to do something like `r := import("rand")` for some reasons). 
2. map structure kinda makes it easier to do something like this:
```golang
importModules := stdlib.GetModules(stdlib.AllModuleNames()...)
delete(importModules, "os")
delete(importModules, "fmt")
importModules["mod1"] = module1                                             // some builtin module
importModules["mod2"] = &objects.SourceModule{Src: src}  // some source module
s.SetImports(importModules)
```

earncef: No, I don't mean within the language but instead in the compiler. Something like this:

```
s.SetImports(map[string]objects.Importable{
    "fmt": stdlib.BuiltinModules["math"],
})
```

which will then mean
```
fmt := import("fmt") // imported math
```

While yes, within the language itself an import can be named anything, at least on the script level this can be easily avoided. Technically making it a map on `script.SetImports` and `compiler.NewCompiler()` is unnecessary (the map key is already within `objects.Importable`).

I do see your point with maps though. Just not sure it's worth the possible scenario described above. I tend to lean towards the cautious approach. That is, actively selecting the modules to include rather than the other way around. It tends to avoid scenarios where a module included in the future would get automatically imported when it shouldn't.

earncef: In any case, for now I suppose disallowing empty module names is probably enough.

PR #154, opened 2019-03-20
fix bytecode encoding/decoding of builtin modules

d5:
This PR is to fix an issue where decoded bytecode from binary has `nil` user functions. Bytecode decoder detects builtin modules (immutable maps with special field `__module_name__`) and replace them with known runtime builtin modules on the fly.

### DIFF
--- a/stdlib/stdlib.go
+++ b/stdlib/stdlib.go
@@ -18,14 +18,15 @@ func AllModuleNames() []string {
 
 // GetModules returns the modules for the given names.
 // Duplicate names and invalid names are ignore.
-func GetModules(names ...string) map[string]objects.Importable {
-	modules := make(map[string]objects.Importable)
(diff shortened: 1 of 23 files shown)


stdlib/stdlib.go
```diff
@@ -18,14 +18,15 @@ func AllModuleNames() []string {
 
 // GetModules returns the modules for the given names.
 // Duplicate names and invalid names are ignore.
-func GetModules(names ...string) map[string]objects.Importable {
-	modules := make(map[string]objects.Importable)
+func GetModules(names ...string) *objects.ModuleMap {
+	modules := objects.NewModuleMap()
```
earncef: Might be better named GetModuleMap

earncef: Also the comment for this function is better modified to reflect that as well given that modules themselves (builtin & source) have different types and the returning object is a ModuleMap

PR #166, opened 2019-03-27
'enum' module

d5:
- [x] first set of enum functions
- [x] add 'is_iterable' builtin functions
- [x] add bytes iterator (now bytes is iterable type)
- [x] update documentation


objects/builtin_type_checks.go
```diff
@@ -181,3 +181,16 @@ func builtinIsCallable(args ...Object) (Object, error) {
 
 	return FalseValue, nil
 }
+
+func builtinIsIterable(args ...Object) (Object, error) {
+	if len(args) != 1 {
+		return nil, ErrWrongNumArguments
+	}
+
+	switch args[0].(type) {
```
earncef: What about custom iterable objects? How about something like:

```
	if _, ok := args[0].(Iterable); ok {
		return TrueValue, nil
	}
```

PR #168, opened 2019-03-31
fix a bug in dead code elimination

d5:
- [x] add some tests

### DIFF
--- a/compiler/compiler.go
+++ b/compiler/compiler.go
@@ -6,7 +6,6 @@ import (
 	"io/ioutil"
 	"path/filepath"
 	"reflect"
-	"sort"
 	"strings"
 
 	"github.com/d5/tengo"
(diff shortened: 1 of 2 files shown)


compiler/compiler.go
```diff
@@ -722,16 +721,15 @@ func (c *Compiler) optimizeFunc(node ast.Node) {
 	// are considered as unreachable.
 
 	// pass 1. identify all jump destinations
-	var dsts []int
+	dsts := make(map[int]bool)
```
d5: bug 1. fail to catch when the same jump destination occurred more than once.

compiler/compiler.go
```diff
@@ -761,7 +759,7 @@ func (c *Compiler) optimizeFunc(node ast.Node) {
 	// pass 3. update jump positions
 	var lastOp Opcode
 	var appendReturn bool
-	endPos := len(newInsts)
+	endPos := len(c.scopes[c.scopeIndex].instructions)
```
d5: bug 2. end position should be from the previous instructions

PR #170, opened 2019-04-01
fix script timeout bug

d5:
- [x] fix an edge case where script timeout is too short and set even before VM starts execution

### DIFF
--- a/runtime/vm.go
+++ b/runtime/vm.go
@@ -1002,10 +1002,9 @@ func (v *VM) run() {
 	}
 }
 
-// StackObjects returns a slice of objects in the stack.
-// This function exists for the tests, and, should not be used to access stack objects.
-func (v *VM) StackObjects() []objects.Object {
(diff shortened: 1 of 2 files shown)


runtime/vm.go
```diff
@@ -1008,9 +1002,10 @@ func (v *VM) run() {
 	}
 }
 
-// Globals returns the global variables.
-func (v *VM) Globals() []objects.Object {
-	return v.globals
+// StackObjects returns a slice of objects in the stack.
+// This function exists for the tests, and, should not be used to access stack objects.
+func (v *VM) StackObjects() []objects.Object {
```
earncef: I'm not sure this is a good idea. Even for tests, there should be no real reason to actually access the contents of the stack. It might be better to do something like:

```
func (v *VM) IsStackEmpty() bool {
    return v.sp > 0
}
```

PR #180, opened 2019-04-08
Substring function

earncef:
String slicing with range doesn't work for variables that are expected to be string but turn out to not be strings (especially when retrieved from json). Substring function deals with such scenarios better than string conversion followed by slicing (works similar to other functions in the text library by first converting the object to string).


earncef:
All done

PR #181, opened 2019-04-09
Fix %v formatting

### LINKED ISSUE #143: proper handling of '%v' of builtin function printf and sprintf
```golang
m := {a: {b: {c: [1, 2, 3]}}}
printf("m: %v", m)
```

This code currently prints `m: map[a:map[b:map[c:[1 2 3]]]]` but it's supposed to be `m: {a: {b: {c: [1, 2, 3]}}}`.

earncef:
fixes #143 

All Objects implement String(), so %v would work automatically if they're not converted to interface.

### DIFF
--- a/stdlib/fmt_test.go
+++ b/stdlib/fmt_test.go
@@ -6,7 +6,8 @@ func TestFmtSprintf(t *testing.T) {
 	module(t, `fmt`).call("sprintf", "").expect("")
 	module(t, `fmt`).call("sprintf", "foo").expect("foo")
 	module(t, `fmt`).call("sprintf", `foo %d %v %s`, 1, 2, "bar").expect("foo 1 2 bar")
-	module(t, `fmt`).call("sprintf", "foo %v", `[1, "bar", true]`).expect(`foo [1, "bar", true]`)
-	module(t, `fmt`).call("sprintf", "foo %v %d", `[1, "bar", true]`, 19).expect(`foo [1, "bar", true] 19`)
-	module(t, `fmt`).call("sprintf", "foo %v", `{a: {b: {c: [1, 2, 3]}}}`).expect(`foo {a: {b: {c: [1, 2, 3]}}}`)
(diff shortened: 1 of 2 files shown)


stdlib/fmt_test.go
```diff
@@ -8,4 +8,5 @@ func TestFmtSprintf(t *testing.T) {
 	module(t, `fmt`).call("sprintf", `foo %d %v %s`, 1, 2, "bar").expect("foo 1 2 bar")
 	module(t, `fmt`).call("sprintf", "foo %v", `[1, "bar", true]`).expect(`foo [1, "bar", true]`)
 	module(t, `fmt`).call("sprintf", "foo %v %d", `[1, "bar", true]`, 19).expect(`foo [1, "bar", true] 19`)
+	module(t, `fmt`).call("sprintf", "foo %v", `{a: {b: {c: [1, 2, 3]}}}`).expect(`foo {a: {b: {c: [1, 2, 3]}}}`)
```
d5: Tests should be like this:

```golang 
module(t, `fmt`).call("sprintf", "").expect("")
module(t, `fmt`).call("sprintf", "foo").expect("foo")
module(t, `fmt`).call("sprintf", `foo %d %v %s`, 1, 2, "bar").expect("foo 1 2 bar")
module(t, `fmt`).call("sprintf", "foo %v", ARR{1, "bar", true}).expect(`foo [1, "bar", true]`)
module(t, `fmt`).call("sprintf", "foo %v %d", ARR{1, "bar", true}, 19).expect(`foo [1, "bar", true] 19`)
module(t, `fmt`).call("sprintf", "foo %v", MAP{"a": IMAP{"b": IMAP{"c": ARR{1, 2, 3}}}}).expect(`foo {a: {b: {c: [1, 2, 3]}}}`)
module(t, `fmt`).call("sprintf", "%v", IARR{1, IARR{2, IARR{3, 4}}}).expect(`[1, [2, [3, 4]]]`)
```

I know my tests were wrong in the first place. :)

earncef: Oh right haha. I’ll fix it.

earncef: All done.

PR #183, opened 2019-04-10
text.trim, text.pad_left and text.pad_right

earncef:


stdlib/text.go
```diff
@@ -440,6 +443,116 @@ func textSubstring(args ...objects.Object) (ret objects.Object, err error) {
...
+			}
+			return
+		}
+	}
+
+	sLen := len(s1)
+	padStrLen := len(s3)
+
+	if sLen >= i2 || padStrLen == 0 {
+		ret = &objects.String{Value: s1}
```
d5: I think we can just return `args[0]`?

d5:
Please update documentation too :)

stdlib/text.go
earncef: args[0] might not actually be a string.

d5: You're right. Haha.

earncef:
All done

Issue #123, opened 2019-02-28
memory safe implementations for print, printf, sprintf


d5:
I'd like to add memory-safe implementation of builtin functions `print`, `printf`, and `sprintf`, and, let the user choose between regular (unchecked) version and memory-safe (checked but slower) version based on their requirements.

PR #184, opened 2019-04-10
Added format (alias for fmt.sprintf) to builtins.

earncef:
String formatting is a very common application in data transformations, and should be available outside of fmt, both in general and in tengomin. 

This is also a precursor to removing support for `string + non-string`. Right now there is an inconsistency in results of of the `+` operator purely depending on whether the first operand is a `string` or not, which is a confusing scenario and can lead to easily avoidable errors. 

It's best to remove support for `string + non-string` and force the user to either forcibly cast every variable to string and append them or to use the `format()` function.

### DIFF
--- a/objects/builtin_format.go
+++ b/objects/builtin_format.go
@@ -1,0 +1,44 @@
+package objects
+
+import (
+	"fmt"
+
+	"github.com/d5/tengo"
+)
+
+func builtinFormat(args ...Object) (Object, error) {
+	numArgs := len(args)
+	if numArgs == 0 {
+		return nil, ErrWrongNumArguments
+	}
+
+	format, ok := args[0].(*String)
+	if !ok {
+		return nil, ErrInvalidArgumentType{
+			Name:     "format",
+			Expected: "string",
+			Found:    args[0].TypeName(),
+		}
+	}
+	if numArgs == 1 {
+		return format, nil // okay to return 'format' directly as String is immutable
+	}
+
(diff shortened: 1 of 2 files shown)


objects/builtin_format.go
```diff
@@ -0,0 +1,44 @@
...
+			Name:     "format",
+			Expected: "string",
+			Found:    args[0].TypeName(),
+		}
+	}
+	if numArgs == 1 {
+		return format, nil // okay to return 'format' directly as String is immutable
+	}
+
+	formatArgs := make([]interface{}, numArgs-1, numArgs-1)
```
golangcibot: S1019: should use make([]interface{}, numArgs - 1) instead (from `gosimple`)

objects/builtin_format.go
```diff
@@ -0,0 +1,44 @@
...
+		case *Int, *Float, *Bool, *Char, *String, *Bytes:
+			formatArgs[idx] = ToInterface(arg)
+		default:
+			formatArgs[idx] = arg
+		}
+	}
+
+	s := fmt.Sprintf(format.Value, formatArgs...)
+
+	if len(s) > tengo.MaxStringLen {
```
d5: I know this is how we do in `fmt.Sprintf`, but, ideally the length of the outcome string has to be checked before it's actually created (`s` in this case). See #123. `fmt` module can be excluded if needed, but, the builtin function cannot be excluded. We need to figure out a way to pre-compute the length and check against the limit.

PR #189, opened 2019-04-23
variable-argument functions

mccolljr:
This is just an initial implementation of variable-argument functions. I'm expecting to need to make changes, but I wanted to demonstrate the initial approach and get feedback before going further.

With this implementation, there are two types of functions supported: regular (`func(a,b,c) {}`) and variadic (`func(...args) {}`). Currently, if a function is variadic, it must have exactly one argument. The VM is responsible for consolidating passed parameters when it processes an `OpCall` instruction on a `CompiledFunction` or `Closure` that is variadic. It does this by consolidating the arguments on the stack into an array, which is then stored on the stack in place of the original first argument. If no arguments are passed to a variadic function call, an empty array is added to the stack as the first argument.

### DIFF
--- a/compiler/parser/parser.go
+++ b/compiler/parser/parser.go
@@ -550,7 +550,7 @@ func (p *Parser) parseFuncType() *ast.FuncType {
 	}
 
 	pos := p.expect(token.Func)
-	params := p.parseIdentList(true)
+	params := p.parseIdentList()
 
 	return &ast.FuncType{
 		FuncPos: pos,
@@ -603,7 +603,7 @@ func (p *Parser) parseIdent() *ast.Ident {
 	}
 }
 
-func (p *Parser) parseIdentList(allowVarArgs bool) *ast.IdentList {
+func (p *Parser) parseIdentList() *ast.IdentList {
 	if p.trace {
 		defer un(trace(p, "IdentList"))
 	}
@@ -613,10 +613,6 @@ func (p *Parser) parseIdentList(allowVarArgs bool) *ast.IdentList {
 	isVarArgs := false
 	if p.token != token.RParen {
 		if p.token == token.Ellipsis {
-			if !allowVarArgs {
-				p.error(p.pos, "variable arguments are not permitted")
-			}
-
 			isVarArgs = true
 			p.next()
 		}
@@ -625,19 +621,11 @@ func (p *Parser) parseIdentList(allowVarArgs bool) *ast.IdentList {
 		for !isVarArgs && p.token == token.Comma {
 			p.next()
 			if p.token == token.Ellipsis {
-				if !allowVarArgs {
-					p.error(p.pos, "variable arguments are not permitted")
-				}
-
 				isVarArgs = true
 				p.next()
 			}
 			params = append(params, p.parseIdent())
 		}
-	}
-
-	if p.token == token.Comma {
-		p.error(p.pos, "variable argument must be last in list")
 	}
 
 	rparen := p.expect(token.RParen)
--- a/compiler/parser/parser.go
+++ b/compiler/parser/parser.go
@@ -550,7 +550,7 @@ func (p *Parser) parseFuncType() *ast.FuncType {
 	}
 
 	pos := p.expect(token.Func)
-	params := p.parseIdentList()
+	params := p.parseIdentList(true)
 
 	return &ast.FuncType{
 		FuncPos: pos,
@@ -603,26 +603,39 @@ func (p *Parser) parseIdent() *ast.Ident {
 	}
 }
 
-func (p *Parser) parseIdentList() *ast.IdentList {
+func (p *Parser) parseIdentList(allowVarArgs bool) *ast.IdentList {
 	if p.trace {
 		defer un(trace(p, "IdentList"))
 	}
 
 	var params []*ast.Ident
 	lparen := p.expect(token.LParen)
+	isVarArgs := false
 	if p.token != token.RParen {
+		if p.token == token.Ellipsis {
+			if !allowVarArgs {
+				p.error(p.pos, "variable arguments are not permitted")
+			}
+
+			isVarArgs = true
+		}
+		p.next()
+
 		params = append(params, p.parseIdent())
-		for p.token == token.Comma {
-			p.next()
-			params = append(params, p.parseIdent())
+		if !isVarArgs {
+			for p.token == token.Comma {
+				p.next()
+				params = append(params, p.parseIdent())
+			}
 		}
 	}
 	rparen := p.expect(token.RParen)
 
 	return &ast.IdentList{
-		LParen: lparen,
-		RParen: rparen,
-		List:   params,
+		LParen:  lparen,
+		RParen:  rparen,
+		VarArgs: isVarArgs,
+		List:    params,
 	}
 }
 
--- a/compiler/parser/parser.go
+++ b/compiler/parser/parser.go
@@ -618,8 +618,8 @@ func (p *Parser) parseIdentList(allowVarArgs bool) *ast.IdentList {
 			}
 
 			isVarArgs = true
+			p.next()
 		}
-		p.next()
 
 		params = append(params, p.parseIdent())
 		if !isVarArgs {
--- a/compiler/parser/parser.go
+++ b/compiler/parser/parser.go
@@ -622,13 +622,24 @@ func (p *Parser) parseIdentList(allowVarArgs bool) *ast.IdentList {
 		}
 
 		params = append(params, p.parseIdent())
-		if !isVarArgs {
-			for p.token == token.Comma {
+		for !isVarArgs && p.token == token.Comma {
+			p.next()
+			if p.token == token.Ellipsis {
+				if !allowVarArgs {
+					p.error(p.pos, "variable arguments are not permitted")
+				}
+
+				isVarArgs = true
 				p.next()
-				params = append(params, p.parseIdent())
 			}
+			params = append(params, p.parseIdent())
 		}
 	}
+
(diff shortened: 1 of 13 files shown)


d5:
Thanks for the PR. I will take a look!

d5:
@mccolljr I will invite you to a Slack channel if you send me an email: me@daniel.gs

runtime/vm_function_test.go
```diff
@@ -12,6 +12,14 @@ func TestFunction(t *testing.T) {
 	expect(t, `f1 := func() {}; f2 := func() { return f1(); }; f1(); out = f2();`, nil, objects.UndefinedValue)
 	expect(t, `f := func(x) { x; }; out = f(5);`, nil, objects.UndefinedValue)
 
+	expect(t, `f := func(...x) { return x; }; out = f(1,2,3);`, nil, &objects.Array{Value: []objects.Object{
```
earncef: It should be possible to have functions with multiple params with the last parameter being variadic.

For example `func(a, ...b)` should be valid.

earncef:
The other concern I have is of performance. I modified the two fib tail call functions in `cmd/bench`. The variadic implementation is quite slow in both cases. I'm assuming it adds a significant overhead when the arguments are consolidated into an array and pushed to the stack. 

The modifications and results:

```
fib := func(...x) {
	if x[0] == 0 {
		return 0 + x[1]
	} else if x[0] == 1 {
		return 1 + x[1]
	}

	return fib(x[0]-1, fib(x[0]-2, x[1]))
}
```

```
fib := func(...x) {
	if x[0] == 0 {
		return x[1]
	} else if x[0] == 1 {
		return x[2]
	}

	return fib(x[0]-1, x[2], x[1]+x[2])
}
```

Results:

BEFORE:
```
fibonacci(35) (tail-call #1)
Result:  9227465
Go:      79.25467ms
Parser:  23.245µs
Compile: 99.047µs
VM:      3.026240441s
fibonacci(35) (tail-call #2)
Result:  9227465
Go:      442ns
Parser:  28.094µs
Compile: 85.925µs
VM:      16.416µs
```

AFTER:

```
fibonacci(35) (tail-call #1)
Result:  9227465
Go:      74.706679ms
Parser:  59.182µs
Compile: 106.193µs
VM:      8.298983466s
fibonacci(35) (tail-call #2)
Result:  9227465
Go:      282ns
Parser:  36.099µs
Compile: 107.335µs
VM:      42.433µs
```

runtime/vm_function_test.go
mccolljr: I fully agree, just wanted to get feedback on the overall approach before I added more complexity. Once we hash out the performance concerns I'll make this change. Should only be a matter of consolidating the last X variables instead of all of them. Parsing-wise, shouldn't add too much complexity. All the magic happens in the `compiler`'s `parseIdentList` function

mccolljr:
@earncef that performance hit is pretty big... but I have to wonder how likely it is that tail-call recursion is going to be used in conjunction with variadic functions. I tried to keep changes to the core structure of the language to a minimum, which meant that the only place I could get information about the callee was in at the vm level, when an `OpCall` instruction is processed. That's why the process of consolidating arguments into an array for variadic functions happens at the vm level. I'm open to suggestions or a nudge in the right direction but I'm not sure what the best way to improve the situation is. Maybe there's some special handling that can be done in the tail call check to update indexes in the argument array rather than pushing the arguments back on the stack? I'l look into it when I have some time later today, but advice would be appreciated.

runtime/vm.go
```diff
@@ -642,6 +642,25 @@ func (v *VM) run() {
 
 			switch callee := value.(type) {
 			case *objects.Closure:
+				// consolidate args into an array
+				if callee.Fn.VarArgs {
+					passedArgs := numArgs
```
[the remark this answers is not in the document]
mccolljr: I made these changes as part of the most recent push. Thanks for simplifying - this is much cleaner.

runtime/vm.go
```diff
@@ -674,6 +693,25 @@ func (v *VM) run() {
 				v.sp = v.sp - numArgs + callee.Fn.NumLocals
 
 			case *objects.CompiledFunction:
+				// consolidate args into an array
+				if callee.VarArgs {
+					passedArgs := numArgs
```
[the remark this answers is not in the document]
mccolljr: I made these changes as part of the most recent push. Thanks for simplifying - this is much cleaner.

runtime/vm_function_test.go
```diff
@@ -12,6 +12,14 @@ func TestFunction(t *testing.T) {
...
+		&objects.Int{Value: 1},
+		&objects.Int{Value: 2},
+		&objects.Int{Value: 3},
+	}})
+
+	expect(t, `f := func(...x) { return x; }; out = f();`, nil, &objects.Array{Value: []objects.Object{}})
+
```
earncef: Needs tests for closures

runtime/vm.go
```diff
@@ -674,9 +698,32 @@ func (v *VM) run() {
 				v.sp = v.sp - numArgs + callee.Fn.NumLocals
 
 			case *objects.CompiledFunction:
+				// consolidate args into an array
+				if callee.VarArgs {
+					realArgs := callee.NumParameters - 1
```
earncef: Perhaps rewrite as the following to avoid unnecessary computations at runtime (and similarly for Closures):

```
realArgs := callee.NumParameters - 1
varArgs := numArgs - realArgs
if varArgs < 0 {
	goto wrongNumberOfParametersCompiledFunction
}
sp := v.sp - varArgs
numArgs = realArgs + 1
args := make([]objects.Object, 0, varArgs)

for i := sp; i < v.sp; i++ {
	args = append(args, v.stack[i])
}

v.stack[sp] = &objects.Array{Value: args}
v.sp = sp + 1
```

compiler/parser/parser.go
```diff
@@ -603,26 +603,50 @@ func (p *Parser) parseIdent() *ast.Ident {
 	}
 }
 
-func (p *Parser) parseIdentList() *ast.IdentList {
+func (p *Parser) parseIdentList(allowVarArgs bool) *ast.IdentList {
```
d5: I guess `allowVarArgs` is always `true` in the current implementation.

d5: So this can be reduced to

```golang
func (p *Parser) parseIdentList() *ast.IdentList {
	if p.trace {
		defer un(trace(p, "IdentList"))
	}

	var params []*ast.Ident
	lparen := p.expect(token.LParen)
	isVarArgs := false
	if p.token != token.RParen {
		if p.token == token.Ellipsis {
			isVarArgs = true
			p.next()
		}

		params = append(params, p.parseIdent())
		for !isVarArgs && p.token == token.Comma {
			p.next()
			if p.token == token.Ellipsis {
				isVarArgs = true
				p.next()
			}
			params = append(params, p.parseIdent())
		}
	}

	rparen := p.expect(token.RParen)

	return &ast.IdentList{
		LParen:  lparen,
		RParen:  rparen,
		VarArgs: isVarArgs,
		List:    params,
	}
}
```

runtime/vm_function_test.go
```diff
@@ -12,6 +12,53 @@ func TestFunction(t *testing.T) {
 	expect(t, `f1 := func() {}; f2 := func() { return f1(); }; f1(); out = f2();`, nil, objects.UndefinedValue)
 	expect(t, `f := func(x) { x; }; out = f(5);`, nil, objects.UndefinedValue)
 
+	expect(t, `f := func(...x) { return x; }; out = f(1,2,3);`, nil, &objects.Array{Value: []objects.Object{
+		&objects.Int{Value: 1},
+		&objects.Int{Value: 2},
+		&objects.Int{Value: 3},
+	}})
+
+	expect(t, `f := func(a, b, ...x) { return [a, b, x]; }; out = f(8,9,1,2,3);`, nil, &objects.Array{Value: []objects.Object{
```
d5: you can do 

```golang
expect(t, `f := func(...x) { return x; }; out = f(1,2,3);`, nil, ARR{1, 2, 3})
```

runtime/vm.go
```diff
@@ -642,9 +642,34 @@ func (v *VM) run() {
...
+					spStart := v.sp - varArgs
+					for i := spStart; i < v.sp; i++ {
+						args = append(args, v.stack[i])
+					}
+
+					v.stack[spStart] = &objects.Array{Value: args}
+					v.sp = spStart + 1
+				}
+
+			wrongNumberOfParametersClosure:
```
d5: we can remove `goto` and don't need `append` for `varArgs`:

```golang
if callee.Fn.VarArgs {
	// if the closure is variadic,
	// roll up all variadic parameters into an array
	realArgs := callee.Fn.NumParameters - 1
	varArgs := numArgs - realArgs
	if varArgs >= 0 {
		numArgs = realArgs + 1
		args := make([]objects.Object, varArgs)
		spStart := v.sp - varArgs
		for i := spStart; i < v.sp; i++ {
			args[i-spStart] = v.stack[i]
		}
		v.stack[spStart] = &objects.Array{Value: args}
		v.sp = spStart + 1
	}
}
```

d5: not sure if this affects benchmark results or not.

mccolljr: agreed, fixed

runtime/vm_function_test.go
mccolljr: ah - i figured there had to be something. Thanks! fixed

compiler/parser/parser.go
mccolljr: yeah, i was being proactive but YAGNI, fixed

PR #191, opened 2019-04-25
Error index check moved out of vm to error object

earncef:


earncef:
Hold off on this one. #192 overrides this.

d5:
@earncef so you want to close this PR?

earncef:
I think given that we're leaving #192 to v2, this can go in as it does give some performance improvements with array access in the meanwhile.

PR #192, opened 2019-04-25
Performance gains by moving IndexGet and IndexSet into Object interface 

earncef:
Removing type assertion to `Indexable` and `IndexAssignable` interfaces improves overall `IndexGet` and `IndexSet` performance significantly at the downside of having to implement these methods for all objects. I think it's a good tradeoff.

BEFORE
```
BenchmarkArrayIndex-28                      5000        298004 ns/op
BenchmarkArrayIndexComparison-28           10000        107746 ns/op
```

AFTER
```
BenchmarkArrayIndex-28                     10000        216165 ns/op
BenchmarkArrayIndexComparison-28           10000        108115 ns/op
```


objects/object_ptr.go
```diff
@@ -39,3 +39,13 @@ func (o *ObjectPtr) IsFalsy() bool {
 func (o *ObjectPtr) Equals(x Object) bool {
 	return o == x
 }
+
+// IndexGet returns an element at a given index.
```
mccolljr: Should an `ObjectPtr` be indexable / index-settable if the underlying object is?

earncef: Tengo doesn't have pointers. `ObjectPtr` is used only internally and should never require indexing.

earncef: #194 fixes #193 so this should be resolved now.

docs/objects.md
```diff
@@ -4,17 +4,15 @@
 
 - [Tengo Objects](#tengo-objects)
   - [Object Interface](#object-interface)
-  - [Callable Interface](#callable-interface)
-  - [Indexable Interface](#indexable-interface)
-  - [Index-Assignable Interface](#index-assignable-interface)
-  - [Iterable Interface](#iterable-interface)
+  - [Callable Objects](#callable-interface)
+  - [Iterable Objects](#iterable-interface)
```
d5: broken link: `#callable-interface` `#iterable-interface`

d5: also should those be under `Object Interface` (one level deeper)? Or, just make them part of `Object` interface explanation, rather than separate sections.

earncef: I've made them part of the Object Interface because ultimately they still rely on the CanItereate and Iterator methods.

PR #195, opened 2019-04-26
explode implementation (for v2)

mccolljr:
This is the PR that relates to the explosion proposal. 

It is currently missing:
- compiler tests
- new syntax documentation
- benchmarks

I can/will create these once the rest of the approach has been validated and performance concerns are addressed.

I opened this so we have a more formal/public way to discuss/review the implementation.
I expect this to be a living, breathing PR for a while while we work through:
1. What features from the proposal we should keep / discard
2. How we want to implement said features

### DIFF
--- a/runtime/vm.go
+++ b/runtime/vm.go
@@ -1,6 +1,7 @@
 package runtime
 
 import (
+	"errors"
 	"fmt"
 	"sync/atomic"
 
@@ -20,6 +21,11 @@ const (
 	// MaxFrames is the maximum number of function frames.
 	MaxFrames = 1024
 )
+
+// internal type used to handle explosion on the stack
+type exploded struct {
+	objects.Object
+}
 
 // VM is a virtual machine that executes the bytecode compiled by Compiler.
 type VM struct {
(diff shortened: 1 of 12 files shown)


runtime/vm.go
```diff
@@ -21,6 +21,11 @@ const (
 	MaxFrames = 1024
 )
 
+// internal type used to handle explosion on the stack
+type exploded struct {
```
earncef: What's the point of this? Seems redundant to me and adds an additional type assertion in `OpArray`, `OpExplode`, and `OpCall`

earncef: Ah, I see what you're doing. However, there should be a better way of doing this without adding an extra type assertion.

earncef: Perhaps create an `Exploded` object that extends `Array`. Then use the `Exploded` object directly.

mccolljr: `OpExplode` wraps the exploded value on the stack in this type to signal to `OpCall` and `OpArray` to handle the explosion.

In `OpArray` I actually use it to short-circuit having to do a bunch of stack modification (like in `OpCall`), so IMO that's not too bad

In `OpCall`, it is a signal to tell the vm to explode an argument onto the stack. I can't just check for `objects.Explodable` since an explodable object may be passed without actually being exploded.

In order to avoid using this `exploded` type, I'd need to change the calling convention for functions to pass the argument count on the stack, and when I went that route it got complex _really_ fast.

Definitely open to other suggestions though.

earncef: The `Exploded` type can be an `Object` that holds the contents of `Explode()` within `OpExplode`. That removes the extra type assertion within `OpArray` and `OpCall`

mccolljr: good idea, done

mccolljr:
closing, as this branch has been merged with the vm_hooks branch

PR #198, opened 2019-05-02
Runtime Hooks and Spread syntax

mccolljr:
- [x] spreads in array literals `[a..., b..., c, d, e...]`
- [x] spreads in function calls `some_func(a..., b, c...)`
- [x] define `RuntimeHooks` interface in objects package
- [x] update `CallableFunc` and `Callable::Call` signatures to take a `RuntimeHooks` value as the first parameter
- [x] `RuntimeHooks` implementation in runtime package
- [x] update `Callable` and `CallableFunc` implementations in standard library
- [x] tests for calling tengo code from go
- [x] tests for spread syntax


stdlib/func_typedefs_test.go
```diff
@@ -492,9 +492,31 @@ func TestFuncAI64RI64(t *testing.T) {
...
+}
+
+func (h dummyHooksVal) Call(val objects.Object, args ...objects.Object) (objects.Object, error) {
+	if h.returnVal == nil {
+		return objects.UndefinedValue, nil
+	}
+	return h.returnVal, nil
+}
+
+type dummyHooksError struct {
```
golangcibot: U1000: type `dummyHooksError` is unused (from `unused`)

mccolljr: yeah yeah...

objects/spread_values.go
```diff
@@ -0,0 +1,42 @@
...
+
+// SpreadValues represents the result of spreading an object.
+type SpreadValues struct {
+	Value []Object
+}
+
+// TypeName returns the name of the type.
+func (o *SpreadValues) TypeName() string { return "spread-values" }
+
+func (o *SpreadValues) String() string { return fmt.Sprintf("<%s>", o.TypeName()) }
```
d5: I think this should actually return an actual string representation something like `[1, 2, 3]...`?

objects/spreadable.go
```diff
@@ -0,0 +1,7 @@
+package objects
+
+// Spreadable represents an object that can be spread
+type Spreadable interface {
```
d5: Unless we want to merge Spreadable into Object, I think we should embed `Object` interface here to force `Spreadable` is always `Object`:

```golang
type Spreadable interface {
   Object

   // Spread should return a slice of Objects
   Spread() []Object
}
```

runtime/vm.go
```diff
@@ -333,7 +336,12 @@ func (v *VM) run() {
 
 			var elements []objects.Object
 			for i := v.sp - numElements; i < v.sp; i++ {
-				elements = append(elements, v.stack[i])
+				elt := v.stack[i]
+				if SpreadValues, ok := elt.(*objects.SpreadValues); ok {
```
d5: should use lowercase like `spreadVal`

runtime/vm.go
```diff
@@ -634,11 +642,36 @@ func (v *VM) run() {
 				v.sp++
 			}
 
+		case compiler.OpSpread:
+			sp := v.sp - 1
+			obj := v.stack[sp]
+			if SpreadValues, ok := obj.(objects.Spreadable); ok {
+				v.stack[sp] = &objects.SpreadValues{Value: SpreadValues.Spread()}
```
d5: we need to track down the number of allocations: `v.allocs`

runtime/vm.go
```diff
@@ -634,11 +642,36 @@ func (v *VM) run() {
...
 			numArgs := int(v.curInsts[v.ip+1])
 			v.ip++
-
-			value := v.stack[v.sp-1-numArgs]
+			spBase := v.sp - 1 - numArgs
+			value := v.stack[spBase]
+
+			// spread args that need it
+			for i := spBase + 1; i < v.sp; i++ {
+				if SpreadValues, ok := v.stack[i].(*objects.SpreadValues); ok {
```
d5: lowercase variable name

stdlib/func_typedefs_test.go
```diff
@@ -492,9 +492,31 @@ func TestFuncAI64RI64(t *testing.T) {
...
+// type dummyHooksError struct {
+// 	err error
+// }
+
+// func (h dummyHooksError) Call(val objects.Object, args ...objects.Object) (objects.Object, error) {
+// 	if h.err == nil {
+// 		return nil, errors.New("panic: no error set in dummyHooksError")
+// 	}
+// 	return nil, h.err
+// }
```
d5: let's delete it if not used

runtime/vm_hooks.go
```diff
@@ -0,0 +1,76 @@
+package runtime
+
+import (
+	"github.com/d5/tengo/compiler"
+	"github.com/d5/tengo/compiler/source"
+	"github.com/d5/tengo/objects"
+)
+
+type hooks struct {
```
d5: can VM itself implement the interface?

```golang
func (v *VM) Call(val objects.Object, args ...objects.Object) (objects.Object, error) {
   // ...
}
```

d5:
we also need to update the documentations before this PR's merged.

objects/spread_values.go
earncef: I think this should return either the Array representation `[1, 2, 3]` or an Array after being spread `1 2 3` (or perhaps with commas. I don't think it should actually add ellipses at the end of the string representation

objects/spread_values.go
```diff
@@ -0,0 +1,42 @@
...
+func (o *SpreadValues) Copy() Object { return &SpreadValues{Value: o.Value} }
+
+// IndexGet is not supported on SpreadValues values
+func (*SpreadValues) IndexGet(Object) (Object, error) {
+	return nil, errors.New("cannot index spread-values")
+}
+
+// IndexSet is not supported on SpreadValues values
+func (*SpreadValues) IndexSet(_, _ Object) error {
+	return errors.New("cannot index spread-values")
```
earncef: Use predefined `ErrNotIndexAssignable`

earncef: or simply extend `ObjectImpl`

mccolljr: extended ObjectImpl, fixed

objects/spread_values.go
```diff
@@ -0,0 +1,42 @@
...
+
+// BinaryOp returns another object that is the result of
+// a given binary operator and a right-hand side object.
+func (o *SpreadValues) BinaryOp(token.Token, Object) (Object, error) { return nil, ErrInvalidOperator }
+
+// Copy returns a copy of the SpreadValues value
+func (o *SpreadValues) Copy() Object { return &SpreadValues{Value: o.Value} }
+
+// IndexGet is not supported on SpreadValues values
+func (*SpreadValues) IndexGet(Object) (Object, error) {
```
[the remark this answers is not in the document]
mccolljr: I updated `SpreadValues` to embed `ObjectImpl`

objects/array.go
```diff
@@ -128,3 +128,6 @@ func (o *Array) Iterate() Iterator {
 		l: len(o.Value),
 	}
 }
+
+// Spread spreads the array into its vaues
```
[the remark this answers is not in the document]
mccolljr: fixed

runtime/vm.go
```diff
@@ -634,11 +642,36 @@ func (v *VM) run() {
 				v.sp++
 			}
 
+		case compiler.OpSpread:
+			sp := v.sp - 1
+			obj := v.stack[sp]
+			if SpreadValues, ok := obj.(objects.Spreadable); ok {
```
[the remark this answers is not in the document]
mccolljr: find-replace refactoring issue, fixed

objects/spread_values.go
```diff
@@ -0,0 +1,42 @@
...
+import (
+	"errors"
+	"fmt"
+
+	"github.com/d5/tengo/compiler/token"
+)
+
+// SpreadValues represents the result of spreading an object.
+type SpreadValues struct {
+	Value []Object
```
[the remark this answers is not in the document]
mccolljr: Can you clarify the question?

objects/spread_values.go
mccolljr: This type should never make it to user code. It's created and then de-structured within the VM into regular object values. I'd rather keep the `String` implementation as it is here to make that clear. It's the same way we handle `String` for `Iterator` values

runtime/vm.go
mccolljr: find-replace refactoring issue, fixed

runtime/vm.go
mccolljr: find-replace refactoring issue, fixed

stdlib/func_typedefs_test.go
mccolljr: will do, i will keep it until I'm sure it's not needed for any additional tests based on this review, though

runtime/vm_hooks.go
mccolljr: I chose this approach because I thought it would be better if end users couldn't type-assert `RuntimeHooks` to `*runtime.VM`, but willing to change if you feel like that's an unnecessary protection

objects/spreadable.go
mccolljr: I merged `Spreadable` into `Object` to keep with the pattern we used for `Callable` and `Indexable`/`IndexSettable`

mccolljr:
> we also need to update the documentations before this PR's merged.

Most recent push updates `tutorial.md` and `objects.md` in the docs folder

runtime/vm.go
mccolljr: would this be all that is needed?
```golang
v.allocs--
if v.allocs == 0 {
    v.err = ErrObjectAllocLimit
    return
}
```

d5:
Looks good. Can you resolve conflicts?

runtime/vm.go
d5: Technically speaking, yes because we're allocating another object. (And, that's why I commented it here.) But, now that I think about it again, from a user standpoint, I'm not sure if it makes sense using `...` operator involves a new object allocation.

mccolljr:
Too many bugs and conflicts. Closing for now, have to do some work to make play nice with current version of `tengo2` branch.

PR #206, opened 2019-05-16
Error handling between Go and Tengo interops

d5:
- Calling Stack
  - **[(0) User]** --_VM.Run_--> 
    - **[(1) VM]** --_OpCall_--> 
      -  **[(2) Go]** --_VM.Call_--> 
          - **[(3) VM]** ->
      - **[(2) Go]** -> 
    - **[(1) VM]** -> 
  - **[(0) User]**

Scenarios:
- [ ] Go function (2) returns an error: VM (1) will treat it as a runtime error and stop its execution.
- [ ] Go function (2) panics: VM (1) will not do anything about it and the panic will be propagated back to the calling user (0).
- [ ] VM (3) returns an error:
    - Go function (2) returns the error (directly or wrapped) back to VM (1), which will treat it as a runtime error and stop its execution.
    - Go function (2) ignores or swallows the error: VM (1) still be able to continue its execution as if nothing happened.
- [ ] VM (3) panics
    - If Go function (2) does not capture (`recover`) or rethrow: VM (1) should not do anything about it, and, the panic should be propagated back to the calling user. 
    - If Go function (2) ignores/swallow the panic and proceeds: VM (1) should not execute anything and propagate the last panic back to the calling user.

This PR is not complete. Test does not pass.

### DIFF
--- a/runtime/vm.go
+++ b/runtime/vm.go
@@ -36,7 +36,6 @@ type VM struct {
 	aborting    int64
 	maxAllocs   int64
 	allocs      int64
-	valid       bool
 	err         error
 }
 
@@ -54,7 +53,6 @@ func NewVM(bytecode *compiler.Bytecode, globals []objects.Object, maxAllocs int6
 		framesIndex: 1,
 		ip:          -1,
 		maxAllocs:   maxAllocs,
-		valid:       true,
 	}
 
 	v.frames[0].fn = bytecode.MainFunction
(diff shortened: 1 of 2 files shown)


runtime/vm.go
```diff
@@ -164,24 +156,20 @@ func (v *VM) Run() (err error) {
 	return nil
 }
 
-func (v *VM) run() (verr error) {
+func (v *VM) run() error {
 	defer func() {
 		if r := recover(); r != nil {
-			v.valid = false
+			v.lastPanic = r
```
earncef: Set v.aborting to 1. I think in the last case the vm will continue past the OpCall if Go swallows the panic and doesn't return an error (even though v.err is not empty)

d5:
Closing this for now. I'll work on error handling and create another PR again.

PR #222, opened 2019-09-01
Correct misattributed `string()` details.

bbuck:


docs/builtins.md
```diff
@@ -50,20 +50,20 @@ type_name("str") // string
...
 
+## string
 
-Optionally it can take the second argument, which will be returned if the first argument cannot be converted to string. Note that the second argument does not have to be string.
+Tries to convert an object to string object. See [Runtime Types](https://github.com/d5/tengo/blob/master/docs/runtime-types.md) for more details on type conversion.
 
 ```golang
-v = string(undefined, "foo")  // v == "foo"
-v = string(undefined, false)  // v == false 
+x := string(123) //  v == "123"
```
geseq: Should be  x == “123”

bbuck: Hey, sorry I never got around to making this change. Thanks for taking care of it. I know it has no effect on the code, but still appreciate the maintenance and effort!

PR #241, opened 2019-12-24
fix internal package issue

### LINKED ISSUE #240: Can not create custom Object since "token" becomes a internal package
The signature of `Object.BinaryOp` relies on then internal package [token](https://github.com/d5/tengo/tree/master/internal/token) which make it hard to build custom Object

```go
BinaryOp(op token.Token, rhs Object) (Object, error)
```

d5:
this should fix #240.


example_test.go
```diff
@@ -0,0 +1,46 @@
+package tengo_test
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/d5/tengo"
+)
+
+func ExampleSimple() {
```
golangcibot: tests: ExampleSimple refers to unknown identifier: Simple (from `govet`)

PR #247, opened 2020-02-12
update vm to raise wrapped runtime errors

ozanh:
special errors returned from custom builtin modules can be identified easily by wrapping and using errors package's Is() and As() method.


ozanh:
Let me know if tests are OK

d5:
looks good to me @geseq  what do you think?

geseq:
Yeah, LGTM.

PR #250, opened 2020-02-16
builtin delete for maps and arrays

ozanh:
builtin `delete` is ready. It works like Go's delete for maps but also works for array types. delete function returns only undefined value or raises runtime error. It mutates given map or array. doc and tests are added.


d5:
Right now Tengo does not support negative indexing into arrays. See https://github.com/d5/tengo/blob/master/objects.go#L223 and https://github.com/d5/tengo/blob/master/objects.go#L873. So I think we should remove it from this PR for consistent behavior.

ozanh:
I will remove it soon you are right I missed it

geseq:
A builtin for map deletes makes perfect sense, but now that I think about it, for arrays, the go syntax of creating slices in the form `a[i:j]` might be better than a delete function. I'm gonna look into this further tomorrow.

ozanh:
Daniel, I thought that for a scripting language such a syntax may not be appealing; `a = a[:2] + a[3:]` if item is in the middle and garbage stays in the slice until the script ends (not a big problem though). Honestly, my future rule engine users, who won't be Gophers, can be confused, I guess.

ozanh:
If tengo will have methods like `Array.pop(index)` or `Array.splice(...)` etc. in the future, delete builtin should not support arrays. There must be only one way to do it.

d5:
Actually I really like the idea of having a separate builtin function of `splice` for array types. That way we can maintain the expected behavior of `delete` (that works for maps) and users can still choose to use `splice` for array manipulations. Thoughts? @geseq @ozanh

ozanh:
It sounds good to me.

geseq:
Yeah splice sounds good.

ozanh:
I will remove array support from delete function if you are OK. Later I am going to implement a splice builtin. Guide me please.

geseq:
Perfect!

PR #254, opened 2020-02-25
builtin splice

### LINKED ISSUE #252: builtin function splice
JS `splice` function looks like [this](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/splice)
```js
let arrDeletedItems = array.splice(start[, deleteCount[, item1[, item2[, ...]]]])
```
I can implement it like this for tengo
```golang
deleted_items := splice(array, start[, delete_count[, item1[, item2[, ...]]]])
```
if `delete_count` is not provided, JS deletes the array items starting from the `start` to the end. If `delete_count` is provided, after deletion (if greater than 0), item1, item2, etc. are inserted to the original array. If `delete_count` is 0 and items are provided then it inserts the items to the array starting from `start` index and nothing deleted from the array. It appends, deletes and inserts with the same function.
What do you think?

ozanh:
closes #252 , closes #250


d5:
1. I guess you tried to copy JS array splice() behavior. If so, shouldn't `splice(arr, start)` delete all items after `start` instead of a single item?
2. Can you put builtin tests in `tengo_test` package?
3. Can you try to keep 80-column? (not a strong requirement but for consistency)

ozanh:
1. You are right, all items must be deleted, I will fix it.
2. OK no problem.
3. OK no problem.

ozanh:
@d5 @geseq  Changes are applied. I have one question about `append` for immutable arrays. After deleting array items, array length and capacity may not be same so can append function change immutable arrays unintentionally, if splice is used before append in code?

d5:
I haven't tested the following code thoroughly, but `splice` should be like:

```golang
	argsLen  := len(args)
	if argsLen == 0 {
		return nil, ErrWrongNumArguments
	}

	array, ok := args[0].(*Array)
	if !ok {
		return nil, ErrInvalidArgumentType{
			Name:     "first",
			Expected: "array",
			Found:    args[0].TypeName(),
		}
	}
	arrayLen := len(array.Value)

	var startIdx int
	if argsLen > 1 {
		arg1, ok := args[1].(*Int)
		if !ok {
			return nil, ErrInvalidArgumentType{
				Name:     "second",
				Expected: "int",
				Found:    args[1].TypeName(),
			}
		}
		startIdx = int(arg1.Value)
		if startIdx < 0 || startIdx >= arrayLen {
			return nil, ErrIndexOutOfBounds
		}
	}

	delCount := len(array.Value)
	if argsLen > 2 {
		arg2, ok := args[2].(*Int)
		if !ok {
			return nil, ErrInvalidArgumentType{
				Name:     "third",
				Expected: "int",
				Found:    args[2].TypeName(),
			}
		}
		delCount = int(arg2.Value)
		if delCount < 0 {
			return nil, ErrIndexOutOfBounds
		}
	}
	// if count of to be deleted items is bigger than expected, truncate it
	if startIdx+delCount > arrayLen {
		delCount = arrayLen - startIdx
	}
	// delete items
	endIdx := startIdx + delCount
	deleted := array.Value[startIdx:endIdx]

	// update items in place
	items := array.Value[:startIdx]
	if argsLen > 3 {
		for i := 3; i < argsLen; i++ {
			items = append(items, args[i])
		}
	}
	array.Value = items

	// return deleted items
	return &Array{Value: deleted}, nil
```

1. minimize redundant conditionals (and avoid potential `nil` referencing)
2. invalid indexing should simply error (to be consistent with other indexing for writes)

what do you think?

ozanh:
Code looks better than mine :) I got lots of errors while testing, tomorrow I will focus on this. Thanks.

ozanh:
@d5  with your implementation I could not handle inserts with or without deletions so I changed the last parts and the updated the tests.

geseq:
LGTM

d5:
yeah looks good to me too!

PR #259, opened 2020-03-15
remove callable args alloc

ozanh:
Removing slice allocation for callable arguments improves speed and allocation if arguments are provided to callable. Benchmark results are provided below. This makes sense for me because I run more than ten thousand of tengo scripts in a second. Let me know if benchmark is wrong or if I missed a point.
[Gist for bechmark](https://gist.github.com/ozanh/368c6383b4e5a90246a829e85217abd2)

Current
```
goos: linux
goarch: amd64
pkg: github.com/d5/tengo/v2
BenchmarkCallAlloc/NoArg        11966798                90.0 ns/op             8 B/op          1 allocs/op
BenchmarkCallAlloc/1Arg          8135835               148 ns/op              24 B/op          2 allocs/op
BenchmarkCallAlloc/2Arg          6741585               177 ns/op              40 B/op          2 allocs/op
```

After removing slice allocation
```
goarch: amd64
pkg: github.com/d5/tengo/v2
BenchmarkCallAlloc/NoArg                14734291                88.1 ns/op             8 B/op          1 allocs/op
BenchmarkCallAlloc/1Arg                 12089188                88.0 ns/op             8 B/op          1 allocs/op
BenchmarkCallAlloc/2Arg                 13995144                89.4 ns/op             8 B/op          1 allocs/op
```

PR #260, opened 2020-03-15
Revert "remove callable args alloc"

d5:
Reverts d5/tengo#259

### DIFF
--- a/vm.go
+++ b/vm.go
@@ -604,7 +604,9 @@ func (v *VM) run() {
 				v.framesIndex++
 				v.sp = v.sp - numArgs + callee.NumLocals
 			} else {
-				ret, e := value.Call(v.stack[v.sp-numArgs : v.sp]...)
+				var args []Object
+				args = append(args, v.stack[v.sp-numArgs:v.sp]...)
+				ret, e := value.Call(args...)
 				v.sp -= numArgs + 1
 
 				// runtime error


d5:
PR #259 could give `Object.Call` implementations a direct access to vm stack slice, which could lead to unexpected behaviors. I'm sure there exist more obvious examples, but the following code

```golang
package main

import (
	"fmt"

	"github.com/d5/tengo/v2"
)


type MyType struct {
	tengo.ObjectImpl
}

func (o *MyType) TypeName() string {
	return "mytype"
}

func (o *MyType) String() string {
	return ""
}

func (o *MyType) Call(args ...tengo.Object) (tengo.Object, error) {
	return &tengo.Array{Value: args}, nil
}

func (o *MyType) CanCall() bool {
	return true
}

func main() {
	s := tengo.NewScript([]byte(`
a := 1
b := 2
out = {
	res: fn(a, b),
	b: b
}
`))
	_ = s.Add("fn", &MyType{})
	_ = s.Add("out", nil)
	c, err := s.Run()
	if err != nil {
		panic(err)
	}
	fmt.Println(c.Get("out").String())
}
```
should output `{res: [1, 2], b: 2}` but with that PR change it outputs `{res: ["b", 2], b: 2}`.

geseq:
You're right. I also looked into other options such like `copy` but they don't offer any measurable improvement in performance. Better to revert #259

ozanh:
As I see Go uses same slice with spread operator. Sorry about that, I didn't know.

d5:
All good. I really appreciate any efforts to improve performance.

PR #263, opened 2020-03-27
Support for #-style comments

### LINKED ISSUE #251: Add shebang support (#!)
To use tengo as scripting language in the terminal, support for shebang (`#!`) would be beneficial.

Additional question, what is the proposed file extension for tengo files?

infinytum:
This pull request allows to use #-style comments in tengo as well as shebang support.  Fixes #251 

Shebang:
```tengo
#!/usr/local/bin/tengo

fmt := import("fmt")
fmt.println("Hello World")
```

Outputs:
```
╰─$ ./test.tengo
Hello World
```

### DIFF
--- a/parser/scanner.go
+++ b/parser/scanner.go
@@ -192,6 +192,18 @@ func (s *Scanner) Scan() (
 			} else {
 				tok = s.switch2(token.Quo, token.QuoAssign)
 			}
+		case '#':
+			if s.ch != '!' {
+				s.ch = '!'
+			}
+			comment := s.scanComment()
+			if s.mode&ScanComments == 0 {
+				// skip comment
+				s.insertSemi = false // newline consumed
+				return s.Scan()
+			}
+			tok = token.Comment
+			literal = comment
 		case '%':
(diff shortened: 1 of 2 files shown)


infinytum:
Good idea. Missed that part ^^"

d5:
Please do not merge this yet. I'm not entirely sure or convinced if we want # comments in Tengo.

infinytum:
Personally I think supporting a wider variety of comment types is nice for people coming from other languages. Especially as this one allows the use of shebang as well.

Would be curious to know your arguments against adding a new comment type though :)

d5:
If the goal is simply to support shebang, it should be implemented in Tengo CLI as a special case, rather than a language-wide comment syntax. Tengo CLI can simply ignore the first line of `.tengo` file if it starts with `#!`.

infinytum:
If thats more to what you'd like to have in Tengo, I can also create a new pull-request that will check for that instead of adding a new comment type.

The goal of this request is to support # comments. And in the same turn it will solve the shebang issue too.

PR #274, opened 2020-05-05
add shebang support

### LINKED ISSUE #251: Add shebang support (#!)

ozanh:
This PR should add a simple Shebang support for Tengo scripts and closes #251

### DIFF
--- a/cmd/tengo/main.go
+++ b/cmd/tengo/main.go
@@ -67,6 +67,7 @@ func main() {
 			os.Exit(1)
 		}
 	} else if filepath.Ext(inputFile) == sourceFileExt {
+		inputData = skipShebang(inputData)
 		err := CompileAndRun(modules, inputData, inputFile)
 		if err != nil {
 			_, _ = fmt.Fprintln(os.Stderr, err.Error())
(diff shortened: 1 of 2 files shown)


ozanh:
Should I add skipShebang function to compile only mode?

d5:
Could be wrong, but I think shebang must be at the beginning of the script (file). I think most tools will assume the same. So I think the code can be simpler?

ozanh:
I checked https://en.wikipedia.org/wiki/Shebang_(Unix) and before `#!` spaces are allowed. I can remove `\r` but spaces and tabs are allowed. For example following works at my Linux machine;

```bash
# bash/sh allow spaces before and after #!
echo -e '\t #! /bin/bash \t\necho works' > test
chmod +x test
./test # works
```

```bash
# python2/3 only allow spaces after #! not before
echo -e '#! \t/usr/bin/python3 \t\nprint("works")' > test.py
chmod +x test.py
./test.py # works
```
Anything after executable's path must also be allowed because command line arguments can be passed to executable as well. I should remove `\r` but not sure about the rest.
How should I proceed?

geseq:
I like the way yaegi does it

ozanh:
I think yaegi copies whole buffer to a new string which may not be necessary.

cmd/tengo/main.go
```diff
@@ -67,6 +67,10 @@ func main() {
 			os.Exit(1)
 		}
 	} else if filepath.Ext(inputFile) == sourceFileExt {
+		if len(inputData) > 1 &&
+			bytes.Compare(inputData[:2], []byte("#!")) == 0 {
+			copy(inputData, "//")
```
d5, line 72: 🤣 this is smart!!!

d5:
Looks good to me!

d5:
@geseq feel free to review and merge

PR #282, opened 2020-05-18
Fix nested file imports

### LINKED ISSUE #281: Nested imports fail
I have three tengo files `uno.tengo`
```
due := import(`due`)

due.funcy(1)
```
the second file `due.tengo`,
```
tre := import(`tre`)


funcy := func(a) {
  return tre.fun(a)
}

due := {
  funcy: funcy
}
export due
```
and the third  `tre.tengo`
```
fun := func(a) {
  return a
}

tre := {
  fun: fun
}
export tre
```

All files are in the same folder. Invoking `tengo uno.tengo` within the same folder throws the error:

> Compile Error: module 'tre' not found
        at due.tengo:1:8

So it seems nested imports don't work.

ozanh:
Closes #281 by enabling child compiler to inherit file import allowance.
Also, a better way to compare bytes for shebang.

### DIFF
--- a/compiler.go
+++ b/compiler.go
@@ -1086,6 +1086,8 @@ func (c *Compiler) fork(
 	child := NewCompiler(file, symbolTable, nil, c.modules, c.trace)
 	child.modulePath = modulePath // module file path
 	child.parent = c              // parent to set to current compiler
+	// Fixes #281 inherit file import from parent compiler
+	child.allowFileImport = c.allowFileImport
 	return child
 }
 
--- a/compiler.go
+++ b/compiler.go
@@ -1086,7 +1086,6 @@ func (c *Compiler) fork(
 	child := NewCompiler(file, symbolTable, nil, c.modules, c.trace)
 	child.modulePath = modulePath // module file path
 	child.parent = c              // parent to set to current compiler
-	// Fixes #281 inherit file import from parent compiler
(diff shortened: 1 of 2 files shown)


compiler.go
```diff
@@ -1086,6 +1086,8 @@ func (c *Compiler) fork(
 	child := NewCompiler(file, symbolTable, nil, c.modules, c.trace)
 	child.modulePath = modulePath // module file path
 	child.parent = c              // parent to set to current compiler
+	// Fixes #281 inherit file import from parent compiler
```
d5, line 1089: I guess this is not needed as it's not an edge case that deserves this note.

d5:
Looks good to me. @geseq want to take a look?

ozanh:
I thought it was important even though I have never needed such feature. Please remove it.

d5:
@ozanh I guess you can remove it because this is your PR? 😄

ozanh:
I was about to sleep when I saw it :) give me some time

geseq:
Looks good otherwise.

idoall:
How to use it? I have the following error here

```
vm.Compile() Error	{"error": "[uno.tengo]Compile Error: module file read error: open due.tengo: no such file or directory\n\tat (main):1:8"
```

PR #285, opened 2020-05-20
enable relative imports

### LINKED ISSUE #280: Enable relative imports
Importing a module `tre.tengo` 
```
fun := func(a) {
  return a
}

tre := {
  fun: fun
}
export tre
```
in another tengo script `due.tengo` (both files are in the same folder) 
```
tre := import(`tre`)

funcy := func(a) {
  return tre.fun(a)
}

due := {
  funcy: funcy
}
export due
```
throws an error, when tengo is invoked as CLI from another directory, e.g. the parent folder :

>Compile Error: module file read error: open tre.tengo: no such file or directory
        at due.tengo:2:8

It works fine however when tengo is invoked from within the folder. This leads me to suspect the import is done relative to the working directory. Relative imports feel more intuitive though, or am I wrong. Is there already a way to import relative to another file?

ozanh:
After considering what could be effective, I came up with this implementation. Docs and some tests are missing and will be added if it is accepted. It closes #280 .
Implementation is based on @d5 [comment](https://github.com/d5/tengo/issues/280#issuecomment-630585317)
Notes:
- `checkCyclicImport` call removed which was called before `compileModule` in `Compiler` for file imports. It is already called in `compileModule`.
- I added testdata folder and bunch of tengo scripts under `cmd/tengo` for testing relative imports with CI (testdata folder may be moved to lib root folder).
- At Go side, user can set import dir to search importable tengo files which is updated for each module compiler based on file's dir.
- Absolute paths are used to ensure detecting cyclic imports.
- Physical file modules can be a symbolic link, they are resolved as well but following symbolic links is limited to max 2 levels (who needs more?).
- Tengo CLI compiles and runs both compiled and module files without any problem.
- I have only Linux machine, I dont know if this implementation works well on windows/mac (TODO).

Comments please @d5 @geseq


compiler.go
```diff
@@ -537,24 +539,31 @@ func (c *Compiler) Compile(node parser.Node) error {
 				moduleName += ".tengo"
 			}
 
+			if !filepath.IsAbs(c.importDir) {
+				if p, err := filepath.Abs(c.importDir); err == nil {
+					c.importDir = p
+				}
+			}
```
d5, line 546: I think `c.importDir` should be resolved to absolute path when it's updated (for example, inside SetImportDir function).

Makefile
```diff
@@ -6,6 +6,8 @@ lint:
 
 test: generate lint
 	go test -race -cover ./...
+	go build ./cmd/tengo
+	./tengo ./cmd/tengo/testdata/test.tengo
```
d5, line 10: I think these test .tengo files can be in `testdata` directory at project root.

cmd/tengo/main.go
```diff
@@ -55,10 +55,34 @@ func main() {
...
+		if compileOutput, err = filepath.Abs(compileOutput); err != nil {
+			fmt.Fprintf(os.Stderr,
+				"Error file path: %s\n", err.Error())
+			os.Exit(1)
+		}
+	}
+
+	inputFile, err = filepath.Abs(inputFile)
+	if err != nil {
+		fmt.Fprintf(os.Stderr, "Error file path: %s\n", err)
```
d5, line 72: `_, _ = fmt.Fprintf`

cmd/tengo/main.go
```diff
@@ -55,10 +55,34 @@ func main() {
...
 	if err != nil {
 		_, _ = fmt.Fprintf(os.Stderr,
-			"Error reading input file: %s", err.Error())
+			"Error reading input file: %s\n", err.Error())
 		os.Exit(1)
 	}
 
+	if compileOutput != "" {
+		if compileOutput, err = filepath.Abs(compileOutput); err != nil {
+			fmt.Fprintf(os.Stderr,
```
d5, line 64: `_, _ = fmt.Fprintf`

cmd/tengo/main.go
```diff
@@ -55,10 +55,34 @@ func main() {
...
+	}
+
+	inputFile, err = filepath.Abs(inputFile)
+	if err != nil {
+		fmt.Fprintf(os.Stderr, "Error file path: %s\n", err)
+		os.Exit(1)
+	}
+
+	if err := os.Chdir(filepath.Dir(inputFile)); err != nil {
+		fmt.Fprintf(os.Stderr,
```
d5, line 77: `_, _ = fmt.Fprintf`

cmd/tengo/main.go
```diff
@@ -55,10 +55,34 @@ func main() {
 	inputData, err := ioutil.ReadFile(inputFile)
 	if err != nil {
 		_, _ = fmt.Fprintf(os.Stderr,
-			"Error reading input file: %s", err.Error())
+			"Error reading input file: %s\n", err.Error())
 		os.Exit(1)
 	}
 
+	if compileOutput != "" {
```
d5, line 62: do we have to resolve `compileOutput` path?

cmd/tengo/main.go
```diff
@@ -55,10 +55,34 @@ func main() {
...
+		}
+	}
+
+	inputFile, err = filepath.Abs(inputFile)
+	if err != nil {
+		fmt.Fprintf(os.Stderr, "Error file path: %s\n", err)
+		os.Exit(1)
+	}
+
+	if err := os.Chdir(filepath.Dir(inputFile)); err != nil {
```
d5, line 76: I think we shouldn't change the working directory to the script file. Import script files using a relative path to the current script file makes sense, but the actual working directory should be the same directory that the user executes the code (either Tengo CLI or an inline Tengo code in Go). That way, if the script calls `os` package functions (e.g. `os.read_file`), they will use the current working directory, not the script file directory. 

In other words, the import path for each module should be used only for importing other modules, but nothing else.

compiler.go
```diff
@@ -956,7 +999,7 @@ func (c *Compiler) checkCyclicImports(
 func (c *Compiler) compileModule(
 	node parser.Node,
 	moduleName, modulePath string,
-	src []byte,
+	src []byte, isFile bool,
```
d5, line 1002: can you put `isFile bool,` in a separate line?

compiler.go
```diff
@@ -634,6 +643,40 @@ func (c *Compiler) EnableFileImport(enable bool) {
 	c.allowFileImport = enable
 }
 
+// SetImportDir sets the initial import directory path for file imports.
+func (c *Compiler) SetImportDir(dir string) {
+	c.importDir = dir
+}
+
+func (c *Compiler) resolveModuleLink(node parser.Node,
```
d5, line 651: can you put parameters in separate lines?
```golang
func (c *Compiler) resolveModuleLink(
	node parser.Node,
	modulePath string, 
	level int,
) (string, error) {
```

compiler.go
```diff
@@ -634,6 +643,40 @@ func (c *Compiler) EnableFileImport(enable bool) {
...
+	modulePath string, level int) (string, error) {
+	if level == 2 {
+		return "", c.errorf(node,
+			"module symbolic link max level is reached with \"%s\"", modulePath)
+	}
+	if info, err := os.Lstat(modulePath); err != nil {
+		return "", c.errorf(node, "module file lstat error: %s in \"%s\"",
+			err.Error(), modulePath)
+	} else if info != nil && info.Mode()&os.ModeSymlink != 0 {
```
d5, line 660: are we following symlinks so we can determine an absolute path of the target module?

compiler.go
ozanh: Yes, it is better.

Makefile
ozanh: Why not

cmd/tengo/main.go
ozanh: Is this for linting?

cmd/tengo/main.go
ozanh: os.Chdir changes working dir below this line; otherwise relative output like this `tengo -o ./a/dir/a/file input.tengo` points to wrong location. Getting absolute path before `os.Chdir` fixes this problem. I am open to suggestions.

compiler.go
ozanh: Ok

compiler.go
ozanh: Ok

compiler.go
ozanh: Yes, I could not resist to implement a solution for symlinks but not a must I guess for now.

cmd/tengo/main.go
ozanh: OK I will not mess with os.Chdir :smile:

compiler.go
ozanh: If user does not call `Script.SetImportDir`, we have an empty `Compiler.importDir`, I should check if it is absolute in compiler in any way, right? I will just change `Script.SetImportDir(string)` to `Script.SetImportDir(string) error`, right?

d5: I think so. RE empty import path, I haven't tested it but I think `filepath.Join("", moduleName)` might work using the current working directory... Please test it.

cmd/tengo/main.go
ozanh: This will be removed after removing `os.Chdir` :+1:

compiler.go
d5: Yeah, if it's only for cyclic dependency detection, let's remove it for now. We can open a separate PR for symlink later.

compiler.go
```diff
@@ -955,8 +996,10 @@ func (c *Compiler) checkCyclicImports(
 
 func (c *Compiler) compileModule(
 	node parser.Node,
-	moduleName, modulePath string,
+	moduleName,
+	modulePath string,
```
d5, line 1000: let's do either 
```golang
moduleName, modulePath string,
```
or
```golang
moduleName string,
modulePath string,
```

d5:
Want to take a look @geseq ?

d5:
@ozanh Let's remove a symlink file in testdata dir too.

geseq:
@d5 I was only able to take a quick look but looks good to me.

d5:
Thanks @geseq. I will merge it after @ozanh removes symlink stuffs.

compiler.go
ozanh: :+1:

compiler.go
ozanh: latter :+1:

Makefile
```diff
@@ -6,6 +6,8 @@ lint:
 
 test: generate lint
 	go test -race -cover ./...
+	go build ./cmd/tengo
+	./tengo ./testdata/cli/test.tengo
```
d5, line 10: I think this should be `go run ./cmd/tengo ./testdata/cli/test.tengo`.

ozanh: Done

ozanh:
> Or, we can add a Compile option for this file resolution behavior, which can be set to the old behavior by default. (And the corresponding CLI flag too.)

In my opinion cli flag is better than new version. Version 3 can have significant changes more important than this.

d5:
Yeah. That's probably better. We will need:
- CLI flag
- Script method
- Compiler method
for this module file resolution behavior.

d5:
Sorry clicked the wrong button.

ozanh:
Actually we already have Script and Compiler methods `SetImportDir()` with this PR, haven't we? If user does use these methods, behavior does not change. Am I missing something?
I definitely need to add a CLI flag though :smile:

d5:
It's not about `SetImportDir`. When `importDir` is not set (default behavior), in the current `master`, file modules are always resolved using the current working directory. But with this PR, they will be resolved relative to the importing script file.

ozanh:
I can change PR to check if `importDir` is empty or not. If empty old behavior, if a dir is set even with a `.` (dot) new behavior is used

compiler.go
```diff
@@ -955,8 +960,10 @@ func (c *Compiler) checkCyclicImports(
 
 func (c *Compiler) compileModule(
 	node parser.Node,
-	moduleName, modulePath string,
+	moduleName string,
+	modulePath string,
```
d5, line 964: looks like we're always passing the same value for `moduleName` and `modulePath`. Let's make it a single parameter `modulePath`.

ozanh:
I had added `initialImportDir` before `isFile` was added so it may not be required and forget about it completely, I will look deeply today and add some more tests.

cmd/tengo/main.go
```diff
@@ -25,13 +25,17 @@ var (
 	compileOutput string
 	showHelp      bool
 	showVersion   bool
-	version       = "dev"
+	// TODO Remove this flag at version 3
+	resolvePath bool
```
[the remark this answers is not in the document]
ozanh: Ok

compiler.go
ozanh: Ok

Issue #265, opened 2020-04-05
spread support


ozanh:
Hi, there was spread syntax support for variadic function calls and arrays in tengo2 branch, which was implemented by @mccolljr #190  . Is there a reason not to implement it in master branch with or without explosions?

tmclane:
Unless I am mistaken that was merged in to v2.

tmclane:
Ok after further digging in to the Git history I can see it was merged into tengo2 branch which doesn't appear to have become the new master.

d5:
I'm open to revisit the feature if anyone wants to try implementing it again, but I didn't merge it to master last time because I found some edge cases that were not covered by the implementation. I really liked @mccolljr 's implementation, but I didn't have more time to identify and/or fix those issues. I figured Tengo should be simple and fast, even if it's missing some nice features. I'm not against the idea of adding new features, but only if they're complete and do not affect the running performance.

ozanh:
@d5 could you please re-open this issue again? I will soon complete my spread implementation. I can share benchmark results and some details here before a PR. My spread implementation will cover spread in arrays, maps and function calls like JS.

ozanh:
I finally added some tests (vm, compiler, parser tests) to the new implementation and a fibonacci benchmark test to see how it affects function call performance. Tests are successful but there may be edge cases that I haven't tested for. Implementation is very similar to ES6 spread operator with little differences. For now, any expression after ellipsis is accepted within array, map, call expressions.
These are from directly vm tests

```go
expectRun(t, `out = [...["a", "b", "c"]]`, nil, ARR{"a", "b", "c"})

expectRun(t, `out = [...["a", "b"], "c"]`, nil, ARR{"a", "b", "c"})

expectRun(t, `
	f1 := func(...a){
		return [3, ...a]
	}
	f2 := func(a, ...b) {
		return f1(...[a, ...b])
	}
	out = f2(...[1, 2])
	`, nil, ARR{3, 1, 2})

expectRun(t, `out = {k2: "v2", ...{k1: "v1"}, ...{k3: 3}}`, nil, k1": "v1", "k2": "v2", "k3": 3})
```
Note: immutable arrays and immutable maps are also spread. Unlike JS, function calls accepts only 1 spread expression which can be the last argument like Go.

I had to introduce a new OpCode which is `OpSpread`. In vm `OpArray OpMap OpCall` sections are updated. As @d5 's performance concerns were noted; New `OpCall` first lines look like this;
```go
[...]

PR #297, opened 2020-05-26
new spread feature

ozanh:
#265 
Docs are still missing but I will add later.
Note: After Ellipsis space is allowed in this implementation not to break compatibility.
Note: Any expression is allowed after ellipsis but it can return an error at parser level but I want to hear your ideas.

### DIFF
--- a/compiler.go
+++ b/compiler.go
@@ -339,14 +339,26 @@ func (c *Compiler) Compile(node parser.Node) error {
 			c.emit(node, parser.OpGetFree, symbol.Index)
 		}
 	case *parser.ArrayLit:
+		spread := 0
 		for _, elem := range node.Elements {
-			if err := c.Compile(elem); err != nil {
+			if expr, ok := elem.(*parser.SpreadExpr); ok {
+				if err := c.Compile(expr); err != nil {
+					return err
+				}
+				spread++
+			} else if err := c.Compile(elem); err != nil {
 				return err
 			}
 		}
-		c.emit(node, parser.OpArray, len(node.Elements))
+		c.emit(node, parser.OpArray, len(node.Elements)+spread)
 	case *parser.MapLit:
 		for _, elt := range node.Elements {
+			if expr, ok := elt.Value.(*parser.SpreadExpr); ok {
+				if err := c.Compile(expr); err != nil {
+					return err
+				}
+				continue
+			}
 			// key
 			if len(elt.Key) > MaxStringLen {
 				return c.error(node, ErrStringLimit)
@@ -496,6 +508,19 @@ func (c *Compiler) Compile(node parser.Node) error {
 				return err
 			}
 			c.emit(node, parser.OpReturn, 1)
+		}
+	case *parser.SpreadExpr:
+		switch node.In {
+		case parser.SpreadInCall, parser.SpreadInMap, parser.SpreadInArr:
+			if node.Expr == nil {
+				return c.errorf(node, "spread expression is nil")
+			}
+			if err := c.Compile(node.Expr); err != nil {
+				return err
+			}
+			c.emit(node, parser.OpSpread, int(node.In))
+		default:
+			return c.errorf(node, "spread (...) not allowed")
 		}
 	case *parser.CallExpr:
 		if err := c.Compile(node.Func); err != nil {
--- a/compiler_test.go
+++ b/compiler_test.go
@@ -305,6 +305,33 @@ func TestCompiler_Compile(t *testing.T) {
 				intObject(2),
 				intObject(3))))
 
+	expectCompile(t, `[...[]]`,
+		bytecode(
+			concatInsts(
+				tengo.MakeInstruction(parser.OpArray, 0),
+				tengo.MakeInstruction(parser.OpSpread, int(parser.SpreadInArr)),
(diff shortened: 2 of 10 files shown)


d5:
Didn't look at the code yet, but a couple of questions on the spread operator (`...`):
1. What's the operator precedence?
2. What does spreading map object do? (Do we even need it?)
3. Syntax-wise, how is it different from [ES6](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Spread_syntax)?

geseq:
I don’t think spreading maps is a good idea. It’d make sense to spread map values (or keys) if maps were ordered like they are in some languages, which is not the case in Go and Tengo.

ozanh:
@geseq spread in function calls operates on only array, spread in array operates on only arrays and spread in map operates on only maps. Spread in map does simply merge maps.

geseq:
Ah I see. That sounds a little inconsistent and confusing to me. More so because spreading within maps when a key already exists in the map like below assumes some sort of ordered action within maps:
`{a:1, ...{a:2}, b:2}`

ozanh:
@geseq latter overwrites the former and other languages do the same while merging objects with spread.

ozanh:
@d5 ellipsis has no precedence it is not a binary expression. In the implementation `parseArrayLit parseMapElementLit parseCall` simply checks the token if it is ellipsis and creates `SpreadExpr` which replaces the next expression.

geseq:
I’m not aware of any language that does this with unordered maps. Correct me if I’m wrong, but if you’re referring to JavaScript objects, I believe they do have have an inherent order to the properties even if it’s not insertion based order.

ozanh:
This implementation respects the insertion order and replaces keys if it is defined in spread map. Current implementation of tengo allows this `a := {b:2, b:4}` and `a == {"b":4}` it is not different.

ozanh:
It is just a syntactic sugar to make appending to arrays, merging maps, function calls easier. It is optional though.

ozanh:
>    2. What does spreading map object do? (Do we even need it?)

It merges maps and probably faster than using a for loop.

>     3. Syntax-wise, how is it different from [ES6](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Spread_syntax)?

@d5 it is similar with ES6 spread but function calls do not accept map, only arrays. I am not sure if there are other edge cases, because of that I created a PR as you requested.

d5:
@ozanh I understand this is not final and work in progress. I asked you to open a PR only because it makes it easier to think about and test some edge cases with the real code. It's not to question/debate certain decisions you have made. Just want to clarify that first.

ozanh:
@d5 Thanks for clarification. I started to implement this for just function calls which is highly required by my application. Then I realized that it could be extended to cover spreading arrays and maps without much effort. I am not a fan of spread though. Early implementations were required to have ellipsis after an expression or literal like Go `f(a, b...)` but using ellipsis before improves readability. Just want to give a background why I needed this.
As tengo still does not have an exception handling, it looks like spread is not a priority for now but why not.

d5:
Gotcha. Why don't we start with the function calls first? `f(a, b...)` 

Spreading in the function call is already familiar to Go developers. Also, we can add array/map expansion later if we want to.

ozanh:
Call arguments are expanded under VM `OpCall` so following is valid with current implementation, different than Go.
```go
f := func(a, ...b) {} // a=1, b=[2, 3]
f(...[1, 2, 3])
```
This throws error.
```go
f := func(a, b) {} // accepts 2 arguments, but 3 arguments in stack.
f(...[1, 2, 3])
```
Spread array items are simply appended to VM stack. It is different than Go because of dynamic nature and it is hard to implement one to one argument-parameter match OR I could not figure it out.
If you are OK with loose argument expansion as above I can keep spread in function call and remove spread in array and map.

Note:

Ellipsis (...) position can be before or after expression, it is up to you. But while coding in `parseMapElementLit()` putting ellipsis token before expression created an opportunity to easily check next expression if it is allowed with ellipsis at parser level. Besides, ES6 spread implementation is like that.  No expression type check is implemented currently to discuss it later with you.

geseq:
I think it‘d be better to keep it consistent with go and have ellipses after the expression, unless there’s a significant runtime improvement in changing it.

d5:
Yeah. I think it's better if it's the same (or similar) as Go syntax.

ozanh:
OK it is easy to check ellipsis token after expression is parsed, not a big deal. What I wonder is what to do with `SpreadIn` type constants `SpreadInCall SpreadInArray SpreadInMap`. They are used as operand for `OpSpread` to distinguish where it is used and also passed to custom types implementing Spreader interface which has only method `Spread(parser.SpreadIn) (Object, error)`. Custom objects can be spread like any other array or map with this method. Please have a look at VM tests and play a little bit before proceeding.

d5:
What I'm proposing is to drop array/map spread operator from the implementation for now. Then we wouldn't need `SpreadIn` type/operands, I guess?

ozanh:
Yes we will not need SpreadIn anymore. I am not sure about the OpSpread operand though. Will we need an operand later for something else? When I have time I will create a new PR by closing this just for expand/spread in function calls.

PR #299, opened 2020-06-01
Spread last call argument

ozanh:
- After feedback from #297 (spread in arrays, maps, call args), I removed spread for array and map, only last call argument can be spread.
- Ellipsis must come after the argument, not before (like Go).
- Only last argument can be spread (like Go).
- Call Args and function params match is not mandatory, `OpCall` takes whatever in the stack after spread and use it as before, no major `OpCall` changes.
- Docs are missing, will be added after PR approval.
- Cost of this is not calculated in terms of performance. Simple  benchmarks #265 show +~1% delta.
- Major version change is not required.
- This is a draft PR.

### DIFF
--- a/compiler.go
+++ b/compiler.go
@@ -497,6 +497,14 @@ func (c *Compiler) Compile(node parser.Node) error {
 			}
 			c.emit(node, parser.OpReturn, 1)
 		}
+	case *parser.SpreadExpr:
+		if node.Expr == nil {
+			return c.errorf(node, "spread expression is nil")
+		}
+		if err := c.Compile(node.Expr); err != nil {
+			return err
+		}
+		c.emit(node, parser.OpSpread)
 	case *parser.CallExpr:
 		if err := c.Compile(node.Func); err != nil {
 			return err
--- a/compiler_test.go
+++ b/compiler_test.go
@@ -483,6 +483,26 @@ func TestCompiler_Compile(t *testing.T) {
 				intObject(3),
 				intObject(0))))
 
+	expectCompile(t, `f1 := func(a) { return a }; f1(...[1, 2]);`,
+		bytecode(
+			concatInsts(
+				tengo.MakeInstruction(parser.OpConstant, 0),
+				tengo.MakeInstruction(parser.OpSetGlobal, 0),
+				tengo.MakeInstruction(parser.OpGetGlobal, 0),
+				tengo.MakeInstruction(parser.OpConstant, 1),
+				tengo.MakeInstruction(parser.OpConstant, 2),
+				tengo.MakeInstruction(parser.OpArray, 2),
+				tengo.MakeInstruction(parser.OpSpread),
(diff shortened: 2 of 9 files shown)


d5:
We would need some forms of array spread in the function call:

Go:

```golang
func f(a ...string) {}

func main() {
	f([]string{"foo", "bar"}...)
}
```

Tengo:

```golang
f := func(...a) {}
f(["foo", "bar"]...)
```

This should be allowed only for the last parameter of function call, and, this is different from the full array spread syntax (e.g. `...[1, 2,3]`) which we don't have to support at the moment.

ozanh:
Current PR only supports spreading expressions only in the function call and that argument can be last argument. Ellipsis must be after expression so this is not valid `...[1, 2, 3]`. Spreading is not supported in array, map literals.
This is valid `f(["foo", "bar"]...)`
Is it correct?

d5:
Yeah. We don't support array/map spread (`...{}` or `...[]`), but we should support `func_call(a, b, [c, d, e]...)` to match Go syntax.

ozanh:
`func_call(a, b, [c, d, e]...)` is supposed to be supported with this PR, have you found a bug or did I miss something?

d5:
No, you're right. It's implemented already. I misread the code. My bad.

d5:
@ozanh not sure if this is still draft or not, but, please change the status when it's ready for review.

ozanh:
I could not find a better way than this implementation. Ready for review :+1:

d5:
I think we should add a flag to `CallExpr` to indicate whether it has an expanding function argument or not, instead of creating a separate `SpreadExpr` and `OpSpread`. They're needed if we're using spread operator elsewhere, but I think they're redundant in the current implementation.

ozanh:
I once implemented a flag at CallExpr but then left it in favor of map and array spreads. But I don't know how to inform OpCall about last argument that will be spread if I remove OpSpread. Can you give a hint ?

d5:
Maybe a second operand to `OpCall`?

ozanh:
I thought a second operand but it breaks backward compatibility, requires version 3. Besides I am OK with second operand. It is silly but what if we use a negative value for first operand if it must be expanded/spread :smile: ? I have labeled tons of types and functions as `Tengo2XXX` so I am avoiding a major upgrade :smile:  but it does not matter if it is the right way.

d5:
I think it's okay because it only breaks the binary compatibility (e.g. someone tries to run the compiled binary file using a new version CLI). Maybe we should clarify this in the release note.

d5:
> It is silly but what if we use a negative value for first operand if it must be expanded/spread 😄 ?

It will decrease the maximum number of arguments.

ozanh:
OK, i will add second operand and value 1 for expansion 0 for old behavior.

ozanh:
Closed due to new PR #302

PR #310, opened 2020-07-20
feat: implements Null and False Coalesce expressions

moisespsena:
Hello!
I implemented FALSE and NULL coalescing operators and assigners.


geseq:
Hi, thanks for the PR. This has been tried before and as @d5 suggested at the time, these are redundant operations. 
Find the discussion here: https://github.com/d5/tengo/pull/79#issuecomment-460954402

If you think this still has a valid use case over the `||` operator, please feel free to comment further.

moisespsena:
## The FALSE COALESCE

Hello, there was no attempt on the operator '||'. The operator is of type FALSE COALESCE and not NULL COALESCE. In this case, my FALSE COALESCE operator is really unnecessary. But what do you think about adding the short form for variable assignment "|| =":

```go
// current
a := ""
a = a || "is empty"
fmt.println(a) // "is empty"

// short form for variable assign
b := false
b ||= "is empty"
fmt.println(a) // "is empty"
```

## The NULL COALESCE

In this case, I didn't find a corresponding implementation:

```go
fmt := import("fmt")

fmt.println(undefined ?? "2") // 2

// false != UNDEFINED
fmt.println(false ?? "1") // false

// short form for variable assign
a := undefined
a ??= "is undefined"
fmt.println(a) // "is undefined"
```
What do you think of this implementation?

moisespsena:
Hi, I rename `?:=` (False Coalesce Assign) to `||=` (LOr Assign) and remove token `?:` (False Coalesce) because it equals to `||` (LOr).

geseq:
I’m not really a fan of the syntax but wouldn’t mind it. @d5

d5:
I'm prioritizing stability over new language features at the moment. Maintenance cost is too high for a couple of part-time contributors.

geseq:
Closing this.

PR #316, opened 2020-09-13
fix local symbol resolution bug

### LINKED ISSUE #314: Function re-assignment in another function causes SIGSEGV in Tengo VM
The following Tengo code:

```golang
func1 := func(a, b) {
    return a * b
}

func2 := func() {
    test := test(1, 2)
}

func2()
```

While completely bonkers (re-assigning a function), produces a SIGSEGV, which should not happen.

```text
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x28 pc=0x52eb39]

goroutine 1 [running]:
github.com/d5/tengo/v2.(*VM).run(0xc000180000)
	/home/runner/work/tengo/tengo/vm.go:542 +0x2ac9
github.com/d5/tengo/v2.(*VM).Run(0xc000180000, 0xc000180000, 0x400)
	/home/runner/work/tengo/tengo/vm.go:77 +0xcf
main.CompileAndRun(0xc0001026f0, 0xc000152000, 0x5e, 0x25e, 0xc0001163d0, 0xc, 0x0, 0xc000104120)
	/home/runner/work/tengo/tengo/cmd/tengo/main.go:144 +0x1d0
main.main()
	/home/runner/work/tengo/tengo/cmd/tengo/main.go:83 +0x420
```

I stumbled upon this by happenstance as a result of a typo in my code.

d5:
Fixes #314.

This fixes a bug where an unassigned local variable is incorrectly referenced in 2 different scenarios:

```golang
(func() {
	fn := fn()
})()
```

This should fail compilation with `unresolved reference` error because `fn` is not yet defined.

```golang
fn := func() {}	
(func() {
    fn := fn()
    return fn
})()
```

This should compile and run correctly. `fn` oh the left-hand side is a new local variable and `fn` on the right-hand side is a global variable defined outside the function.

### DIFF
--- a/vm_test.go
+++ b/vm_test.go
@@ -1796,9 +1796,9 @@ func() {
 	// 'f' in RHS at line 4 must reference global variable 'f'
 	// See https://github.com/d5/tengo/issues/314
 	expectRun(t, `
-f := func() { return 2 }	
+f := func() { return 2 }
 out = (func() {
-    f := f()
+	f := f()
 	return f
 })()
 	`, nil, 2)
--- a/vm_test.go
+++ b/vm_test.go
@@ -1792,6 +1792,16 @@ func() {
 		nil, true)
 	expectRun(t, `out = func(v) { for ;;v++ { if v == 3 { break } } }(1)`,
 		nil, tengo.UndefinedValue)
+
+	// 'f' in RHS at line 4 must reference global variable 'f'
+	// See https://github.com/d5/tengo/issues/314
+	expectRun(t, `
+f := func() { return 2 }	
+out = (func() {
+    f := f()
+	return f
+})()
+	`, nil, 2)
 }
 
 func TestIf(t *testing.T) {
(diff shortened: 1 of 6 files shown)


vm_test.go
```diff
@@ -1792,6 +1792,16 @@ func() {
 		nil, true)
 	expectRun(t, `out = func(v) { for ;;v++ { if v == 3 { break } } }(1)`,
 		nil, tengo.UndefinedValue)
+
+	// 'f' in RHS at line 4 must reference global variable 'f'
+	// See https://github.com/d5/tengo/issues/314
+	expectRun(t, `
+f := func() { return 2 }	
+out = (func() {
+    f := f()
```
geseq, line 1801:
```suggestion
	 f := f()
```

geseq:
Tiny suggestion but otherwise looks good

Issue #326, opened 2021-01-11
 please support range syntax


endink:
array define : [1,2,3,4,5]
range define: [1..5]

This syntax is useful in an inline environment to effectively shorten script lengths.

Some language support it like kotlin, groovy...etc

geseq:
This is actually not a bad idea. @d5 any thoughts on this? I wouldn't mind making some time for this.

d5:
Sure. Can we discuss more details on the syntax first?

geseq:
Just looking up how this is done in other languages now but from a cursory overview I’m leaning towards the python style `range` builtin function rather than changing the array syntax.

d5:
Haha. Yeah I like that!

d5:
So the new builtin function will return an array of ints?

geseq:
Yes, an array of ints from 0 to a `n-1` If one argument is provided, `start` to `end-1` with two args, or `start` to `end-1` with `step` Increment with three args

endink:
I think if it's just a built-in, function, and we already have enough APIs to implement it ourselves, the built-in bracket syntax support would be much easier to read for line expression scenarios

geseq:
I’m not sure I agree that brackets add something to expressiveness that a builtin `range` function doesn’t.

It’s also IMO an unnecessary addition to the syntax which I’d rather leave as is unless there’s a really good reason to change it.

Additionally, the bracket syntax is limited. It doesn’t easily allow you to add a step size.

endink:
I use Tengo in a database sharding proxy project, it allows me to simplify some definitions, such as database node definitions: DS_1, DS_2, DS_3. If I have this syntax I can say DS_{[1..3]}, this makes the configuration file simple and easy to read. If it's a function: DS_{rang(1,3)}, I don't think it's as clear so.

In my project, the more complex 
inline expression:
ds_{[1..9]}.table_{[1..5]}

This expression is in the configuration file, so avoiding the function is easier to read.

endink:
BTW: Because missing this syntax, I implemented a "range" function on my project, and this issue comes from a feeling that the configuration file is ugly.

geseq:
I suppose we can achieve something like that with `[start:step:end]` and `[start:end]` syntax but I’m not convinced of the need for it. While your example expression does not look pretty it is readable. I’m reluctant to change the language syntax to make this pretty.

@d5 thoughts?

d5:
[...]

PR #328, opened 2021-03-02
PR: builtin range function #326

endink:
add builtin range function:   

range(start, stop[, step])   

usage examples: 

range(0,0) ----> []   
range(0, 5) -----> [0, 1, 2, 3, 4]   
range(0, 10, 3) -----> [0, 3, 6, 9]
range(0, -10, 2) -----> [0, -2, -4, -6, -8]

special:

The step must be greater than 0 to prevent infinite loops

### DIFF
--- a/errors.go
+++ b/errors.go
@@ -49,6 +49,10 @@ var (
 	// ErrNotImplemented is an error where an Object has not implemented a
 	// required method.
 	ErrNotImplemented = errors.New("not implemented")
+
+	// ErrNotImplemented is an error where an Object has not implemented a
+	// required method.
+	ErrInvalidRangeStep = errors.New("range step must be greater than 0")
 )
 
 // ErrInvalidArgumentType represents an invalid argument value type error.
--- a/errors.go
+++ b/errors.go
@@ -50,8 +50,7 @@ var (
 	// required method.
 	ErrNotImplemented = errors.New("not implemented")
 
-	// ErrNotImplemented is an error where an Object has not implemented a
-	// required method.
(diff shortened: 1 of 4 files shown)


geseq:
Nice one! 

I would expect range(0,0) to be [] for consistency.

endink:
@geseq Okay, i will change range(0,0) to be [] , any ideas for support for a single parameter? 

range(5) to [0, 1, 2, 3, 4], current implementions are not supported, if you want, i can support it

geseq:
I think you might be right and I can see one argument being confusing. I pretty much pulled the idea out of python but we don’t need to stick exactly to python syntax.

endink:
@geseq  ok, Use the same upper and lower bounds will get empty array ( [] ) as you expect, github action seems to have some problems ? please check this PR. I think it's done now !

geseq (approved):

errors.go
```diff
@@ -49,6 +49,10 @@ var (
 	// ErrNotImplemented is an error where an Object has not implemented a
 	// required method.
 	ErrNotImplemented = errors.New("not implemented")
+
+	// ErrNotImplemented is an error where an Object has not implemented a
```
d5, line 53: please update comment?

geseq:
I've merged a fix for the GitHub Actions to master. Please rebase with master

errors.go
endink: @d5 ok, it fixed

d5:
please fix linting issues too!

endink:
emmmm.... okay, I will fix go lint problem

PR #330, opened 2021-03-29
Introduce builtin function go and makechan

Bai-Yingjie:
Hi,

I've added 2 built-in functions: govm and makechan, to run multiple VMs in a similar way that native Golang's go and channel do.

govm(fn, arg1, arg2, ...) starts a goroutine which run fn(arg1, arg2, ...) in a new VM cloned from the current running VM, and returns a goroutineVM object that has wait, result, abort methods. The parent VM and child VM share the same "space": they see the same variables similar as goroutines share the same address space in native Golang.

```golang
var := 0

f1 := func(a,b) { var = 10; return a+b }
f2 := func(a,b,c) { var = 11; return a+b+c }

gvm1 := govm(f1,1,2)
gvm2 := govm(f2,1,2,5)

fmt.println(gvm1.result()) // 3
fmt.println(gvm2.result()) // 8
fmt.println(var) // 10 or 11
```

But unlike Golang `go` semantic which has no return value, `govm` returns a "control handle", so that's why I think `govm` would be better to understand the behavior.

In order to make it real useable, I also added makechan to send/receive object between different VMs.

A lockless server can be like this:
```golang
sharedReqChan := makechan(128)

client = func(name, interval) {
	print := func(s) {
		fmt.println(name, s)
	}

	print("started")
	msg := {data:"hello", chan:makechan(1)}
	sharedReqChan.send(msg)
	print(msg.chan.recv())

	i := 0
	for {
		msg := [i, makechan(1)]
		sharedReqChan.send(msg)
		rep := msg[1].recv()
		print(rep)
		i++
		times.sleep(interval*times.second)
	}
}

asyncPoolServer := func() {
	print := func(...s) {
		fmt.println("asyncPoolServer: ", s...)
	}
	print("started")

	repChan := makechan(128)
	reqChan := makechan(128)

	handle := func(req) {
		repChan.send([req[0]+100, req[1]])
	}

	responder := func() {
		for {
			rep := repChan.recv()
			rep[1].send(rep[0])
		}
	}

	worker := func(num) {
		for {
			req := reqChan.recv()
			print("worker", num)
			handle(req)
		}
	}

	dispatcher := func() {
		for {
			req := sharedReqChan.recv()
			if is_map(req) {
				print(req.data)
				req.chan.send("world")
			} else {
				reqChan.send(req)
			}
		}
	}

	govm(worker, 1)
	govm(worker, 2)
	govm(worker, 3)
	govm(worker, 4)

	govm(responder)
	govm(dispatcher).wait()
}

for i :=0; i < 20; i++ {
	govm(client, format("client %d: ", i), 1)
}
govm(asyncPoolServer).wait()
```

How do you think about it? I am new here, but I love tengo so much, any comments/advices are welcome~

### DIFF
--- a/builtins.go
+++ b/builtins.go
@@ -1,130 +1,43 @@
 package tengo
 
-var builtinFuncs = []*BuiltinFunction{
-	{
-		Name:  "len",
-		Value: builtinLen,
-	},
-	{
-		Name:  "copy",
-		Value: builtinCopy,
-	},
-	{
-		Name:  "append",
-		Value: builtinAppend,
-	},
-	{
-		Name:  "delete",
-		Value: builtinDelete,
-	},
-	{
-		Name:  "splice",
-		Value: builtinSplice,
-	},
-	{
-		Name:  "string",
-		Value: builtinString,
-	},
-	{
-		Name:  "int",
-		Value: builtinInt,
-	},
-	{
-		Name:  "bool",
-		Value: builtinBool,
-	},
-	{
-		Name:  "float",
-		Value: builtinFloat,
-	},
-	{
-		Name:  "char",
-		Value: builtinChar,
-	},
-	{
-		Name:  "bytes",
-		Value: builtinBytes,
-	},
-	{
-		Name:  "time",
-		Value: builtinTime,
-	},
-	{
-		Name:  "is_int",
-		Value: builtinIsInt,
-	},
-	{
-		Name:  "is_float",
-		Value: builtinIsFloat,
-	},
-	{
-		Name:  "is_string",
-		Value: builtinIsString,
-	},
-	{
-		Name:  "is_bool",
-		Value: builtinIsBool,
-	},
-	{
-		Name:  "is_char",
-		Value: builtinIsChar,
-	},
-	{
-		Name:  "is_bytes",
-		Value: builtinIsBytes,
-	},
-	{
-		Name:  "is_array",
-		Value: builtinIsArray,
-	},
-	{
-		Name:  "is_immutable_array",
-		Value: builtinIsImmutableArray,
-	},
-	{
-		Name:  "is_map",
-		Value: builtinIsMap,
-	},
-	{
-		Name:  "is_immutable_map",
-		Value: builtinIsImmutableMap,
-	},
-	{
-		Name:  "is_iterable",
-		Value: builtinIsIterable,
-	},
-	{
-		Name:  "is_time",
-		Value: builtinIsTime,
-	},
-	{
-		Name:  "is_error",
-		Value: builtinIsError,
-	},
-	{
-		Name:  "is_undefined",
-		Value: builtinIsUndefined,
-	},
-	{
-		Name:  "is_function",
-		Value: builtinIsFunction,
-	},
-	{
-		Name:  "is_callable",
-		Value: builtinIsCallable,
-	},
-	{
-		Name:  "type_name",
-		Value: builtinTypeName,
-	},
-	{
-		Name:  "format",
-		Value: builtinFormat,
-	},
-	{
-		Name:  "range",
-		Value: builtinRange,
-	},
+var builtinFuncs []*BuiltinFunction
+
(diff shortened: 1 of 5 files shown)


d5:
Hello! Thanks for opening a PR. I really like the idea of adding some concurrency constructs to Tengo. I think this is a good start. But, there's one thing that I have to mention before going into more details. It seems that this change can significantly slow down the running performance of the regular non-concurrent code. _(E.g. The Fibonacci benchmark is ~10% slower.)_ Also, I personally prefer to keep Tengo as minimal and simple as possible. Concurrency can significantly increase the maintenance costs overall, which we cannot afford at the moment.

Bai-Yingjie:
Hi d5, 
Thanks for your reply! I will look into the performance drop, right now I have no clue... 
The added feature should only affect the running of the cloned VM but even in that case the costs would be 
* VM creation and wrapper entries in `v.frames[0]`  and `v.stack[0]`
* in the case `parser.OpCall` of `(*VM).run()`, a type assertion is added to check if the BuiltinFunction needs vmObj, but this is in Object.Call path, not CompiledFunction call path...

I will dig deeper of the Fibonacci test to understand why...

Bai-Yingjie:
I have run the Fibonacci bench in cmd/bench several times manually, but the result varies from each run, from master branch I got:
```
$ ./bench                                 
fibonacci(35)                             
Result:  9227465                          
Go:      78.809522ms                      
Parser:  220.84µs                         
Compile: 347.96µs                         
VM:      8.494114298s  // this number in several runs: 7.840718186s, 10.21890762s, 7.81819645s, 9.664891104s, 9.262624859s 
fibonacci(35) (tail-call #1)              
Result:  9227465                          
Go:      129.355829ms                     
Parser:  64.265µs                         
Compile: 268.308µs                        
VM:      8.971564671s // this number in several runs: 7.64307332s, 9.831562385s, 9.898320392s, 9.163809704s, 7.817868341s
fibonacci(35) (tail-call #2)              
Result:  9227465                          
Go:      586ns                            
Parser:  56.161µs                         
Compile: 175.593µs                        
VM:      58.555µs  // this number in several runs: 109.931µs, 31.754µs, 41.383µs, 83.342µs, 42.716µs
```

And with this feature branch, on the same X86 PC, I got:
```
$ ./bench_govm
fibonacci(35)
Result:  9227465
Go:      119.630189ms
Parser:  60.796µs
Compile: 380.181µs
VM:      9.183400119s // this number in several runs: 7.647262488s, 8.710117103s, 8.138667236s, 7.610809091s, 8.748246639s
fibonacci(35) (tail-call #1)
Result:  9227465
Go:      134.411886ms
Parser:  133.77µs
Compile: 365.933µs
VM:      9.972241885s // this number in several runs: 7.123116778s, 9.14368598s, 7.47340407s, 6.671002294s, 11.595864038s
fibonacci(35) (tail-call #2)
Result:  9227465
Go:      740ns
Parser:  69.846µs
Compile: 495.052µs
VM:      45.137µs // this number in several runs: 113.465µs, 82.669µs, 44.878µs, 45.22µs, 46.951µs
```

I know it is just a manual run, but from the result, it looks to me more about the result variation of the same binary in different runs...

I will try to add more concurrency test to verify this goroutineVM feature.

geseq:
As far as I can tell the fib benchmark itself shouldn’t be affected as there’s nothing in the execution path within the vm that has changed. I ran it a few times and it’s mostly inconclusive.

geseq:
While I agree that this significant maintenance cost, I think adding concurrency constructs might be worth it.

Bai-Yingjie:
Thanks @geseq 
Regarding the complexity of adding concurrency VM, fully agree that more concurrency constructs like mutex are needed, which brings more complexity... but worth it.
I will continue to test this go VM feature in my dev branch, to make it more stable.

jvegaseg:
What about this PR. I think it could afford my needs: #335

geseq:
@jvegaseg I think  @Bai-Yingjie is still working on improving the stability of this. We’ll re-evaluate once it’s ready.

PR #337, opened 2021-06-19
implement builtin function len base on Measurable interface

uncle-gua:


d5:
Thanks for opening a PR. Can you update `Length()` to return `int` instead of `*Int`? It can be easily wrapped as `Int` inside VM code.

d5:
Why did you change Bool value? It's intentionally non-public. https://github.com/d5/tengo/blob/master/objects.go#L270-L272

uncle-gua:
> 
> 
> Why did you change Bool value? It's intentionally non-public. https://github.com/d5/tengo/blob/master/objects.go#L270-L272

If non-public, how can i inject bool variable from go host, NewBool?

d5:
You should use `TrueValue` or `FalseValue` (https://github.com/d5/tengo/blob/master/objects.go#L17-L20) instead. Maybe that should have been documented somewhere.

geseq:
`Bool` is not `Measurable` in any case, so it's not a change relevant to this PR.

uncle-gua:
> 
> 
> You should use `TrueValue` or `FalseValue` (https://github.com/d5/tengo/blob/master/objects.go#L17-L20) instead. Maybe that should have been documented somewhere.

type D struct {
	Flag1 bool
	Flag2 bool
	Flag3 bool
}

d := &D{Flag: false}

// =======if bool Value is public, like this===========
func (m *futuresModule) MyFunction(args ...wave.Object) (ret wave.Object, err error) {
    return &wave.Map{
	Value: map[string]wave.Object{
		"flag1":        &wave.Bool{Value: d.Flag1},
		"flag2":        &wave.Bool{Value: d.Flag2},
		"flag3":        &wave.Bool{Value: d.Flag3},
	},
    }, nil
}

// ========or bool Value is not public, will like this===========

flag1 := tengo.FalseValue
if d.Flag1 {
    flag1 := tengo.TrueValue
}

flag2 := tengo.FalseValue
if d.Flag2 {
    flag2 := tengo.TrueValue
}

flag3 := tengo.FalseValue
if d.Flag3 {
    flag3 := tengo.TrueValue
}

flag4 := tengo.FalseValue
if d.Flag4 {
    flag4 := tengo.TrueValue
}

func (m *futuresModule) MyFunction(args ...wave.Object) (ret wave.Object, err error) {
    return &wave.Map{
	Value: map[string]wave.Object{
		"flag1":        flag1 ,
		"flag2":        flag2,
		"flag3":        flag3,
		"flag5":        flag4,
	},
    }, nil
}

uncle-gua:
> 
> 
> `Bool` is not `Measurable` in any case, so it's not a change relevant to this PR.



> 
> 
> You should use `TrueValue` or `FalseValue` (https://github.com/d5/tengo/blob/master/objects.go#L17-L20) instead. Maybe that should have been documented somewhere.

type D struct {
Flag1 bool
Flag2 bool
Flag3 bool
}

d := &D{Flag: false}

// =======if bool Value is public, like this===========
func (m *myModule) MyFunction(args ...wave.Object) (ret wave.Object, err error) {
return &wave.Map{
Value: map[string]wave.Object{
"flag1": &wave.Bool{Value: d.Flag1},
"flag2": &wave.Bool{Value: d.Flag2},
"flag3": &wave.Bool{Value: d.Flag3},
},
}, nil
}

// ========or bool Value is not public, will like this===========
func (m *myModule) MyFunction(args ...wave.Object) (ret wave.Object, err error) {

flag1 := tengo.FalseValue
if d.Flag1 {
flag1 := tengo.TrueValue
}

flag2 := tengo.FalseValue
if d.Flag2 {
flag2 := tengo.TrueValue
}

flag3 := tengo.FalseValue
if d.Flag3 {
flag3 := tengo.TrueValue
}
return &wave.Map{
Value: map[string]wave.Object{
"flag1": flag1 ,
"flag2": flag2,
"flag3": flag3,
},
}, nil
}

geseq:
Not really. Just use something like

```
func boolValue(b bool) *tengo.Bool {
    if b {
        return tengo.TrueValue
    }
    return tengo.FalseValue
}
```

PR #341, opened 2021-07-27
Basic http

0xor1:
adding the basic http client with byte limiting, doesn't allow for streaming requests, but should be good enough for most use cases.
[this](https://github.com/0xor1/tengo/compare/dryer-type-checking...0xor1:http?expand=1) is the matching change on the dry-er arg type checking.

basic usage:
```golang
fmt := import("fmt")
http := import("http")

res := http.do("GET", "http://example.test", {"my-header": "why not"}, bytes("this is the body"))

fmt.println(res.code == 200) // true
fmt.println(res.status == "200 OK") // true
fmt.println(len(res.headers) > 0) // map, true
fmt.println(len(res.body) > 0) // bytes, true
```

### DIFF
--- a/stdlib/builtin_modules.go
+++ b/stdlib/builtin_modules.go
@@ -15,4 +15,5 @@ var BuiltinModules = map[string]map[string]tengo.Object{
 	"json":   jsonModule,
 	"base64": base64Module,
 	"hex":    hexModule,
+	"http":   httpModule,
 }
--- a/stdlib/http.go
+++ b/stdlib/http.go
@@ -1,0 +1,115 @@
+package stdlib
+
+import (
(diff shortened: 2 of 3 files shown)


0xor1:
looks like it's failing because it's using go 1.13, I developed in go 1.16 any reason the project go version isn't keeping with latest go?

geseq:
Unless we’re releasing a new major version I’d prefer this be backwards compatible with older Go versions, at least to a version or two before the current version at the time of the 2.0 release of tengo.

raifpy (commented):
Perfect!

tmm1:
can this be updated and merged?

Issue #344, opened 2021-09-02
Unrecoverable panic from VM when using (c *Compiled) RunContext


ganehag:
I recently had a `panic: runtime error: integer divide by zero` thrown by `github.com/d5/tengo/v2.(*Compiled).RunContext`. My immediate thought was to safeguard my code with a proper recovery clause in a defer statement. Which I did, only to discover that the issue persisted.

> panic: runtime error: integer divide by zero
> 
> goroutine 11 [running]:
> github.com/d5/tengo/v2.(*Int).BinaryOp(0xc000025838, 0xe, {0x84bd18, 0xc000025868})
> 	../github.com/d5/tengo/v2/objects.go:1030 +0xac5
> github.com/d5/tengo/v2.(*VM).run(0xc0001e2000)
> 	../github.com/d5/tengo/v2/vm.go:116 +0x2b8c
> github.com/d5/tengo/v2.(*VM).Run(0xc0001e2000)
> 	../github.com/d5/tengo/v2/vm.go:77 +0xd2
> github.com/d5/tengo/v2.(*Compiled).RunContext.func1()
> 	../github.com/d5/tengo/v2/script.go:222 +0x26
> created by github.com/d5/tengo/v2.(*Compiled).RunContext
> 	../github.com/d5/tengo/v2/script.go:221 +0x2c9

Upon investigating the code in `github.com/d5/tengo/v2/script.go:221`, I realised that a Go function inside RunContext executed the VM. Unfortunately, this Go function was not recovering from the panic thrown by the VM, and my code can't recover from it due to it being raised in a Go function.

Patching the code in `github.com/d5/tengo/v2/script.go:221` so that the Go function uses a defer statement, returning the recovered error from RunContext to the caller;

```go
// github.com/d5/tengo/v2/script.go:221
go func() {
	defer func() {
		if r, ok := recover().(error); ok {
			ch <- r
		}
	}()
	ch <- v.Run()
}()
```

Now, I am not sure if this is the proper way to do this. So someone with a more profound knowledge of Go-ism has to weigh in here.

If this is an acceptable solution, then maybe a similar defer statement should be added to `(c *Compiled) Run()` on L206.

d5:
Not entirely sure what you expect, but if your solution is to recover at `Compiled.Run`, you should be able to achieve the same goal with defer with recover inside the function that calls `Compiled.Run`. Also if your code throws `divide by zero`, you should fix that one first no matter what.

ganehag:
[...]

PR #345, opened 2021-09-02
Handle panics by deferring recover as an error

ganehag:
As described in issue #344. I've added a defer statement to handle panics.

### DIFF
--- a/script.go
+++ b/script.go
@@ -219,6 +219,11 @@ func (c *Compiled) RunContext(ctx context.Context) (err error) {
 	v := NewVM(c.bytecode, c.globals, c.maxAllocs)
 	ch := make(chan error, 1)
 	go func() {
+		defer func() {
+			if r, ok := recover().(error); ok {
+				ch <- r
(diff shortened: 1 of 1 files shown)


script.go
```diff
@@ -219,6 +219,11 @@ func (c *Compiled) RunContext(ctx context.Context) (err error) {
 	v := NewVM(c.bytecode, c.globals, c.maxAllocs)
 	ch := make(chan error, 1)
 	go func() {
+		defer func() {
+			if r, ok := recover().(error); ok {
```
d5, line 223: I guess the recovered value doesn't have to be `error` right? We can wrap it as error if it's not an `error` object.

ganehag: I've changed it to look for string and error with a type switch, defaulting to a prefix of "unknown panic:" if the type is unknown.

geseq (approved):
LGTM and looks like a decent addition as well. @d5?

d5 (approved):

PR #346, opened 2021-09-03
optimize Iterator.

### LINKED ISSUE #196: iterator performance
Currently `for-in` statement is implemented by unpacking it into regular `for` statement and a hidden value `:it` ([here](https://github.com/d5/tengo/blob/master/compiler/compiler_for.go#L80)). It works correctly, but, one problem with this approach is that we do lots of type assertions (to `Iterator`) during the iterations: [OpIteratorNext](https://github.com/d5/tengo/blob/master/compiler/compiler_for.go#L80), [OpIteratorKey](https://github.com/d5/tengo/blob/master/runtime/vm.go#L1031), [OpIteratorValue](https://github.com/d5/tengo/blob/master/runtime/vm.go#L1040). We can improve the running performance by either implementing `for-in` statement completely differently, or, somehow removing `Iterator` type assertions. _(We may be able to merge `Iterable` interface to `Object`, but, it doesn't feel quite right to merge `Iterator` and `Object` interface.)_

wdvxdr1123:
Fixes #196

This PR introduce a iterator stack on VM. which improves the performance of iterator, and allows compiler generate less bytecode.


d5:
I will take a deeper look later, but overall, it looks good to me. @geseq thoughts?

geseq:
Nice idea! I’ll take a deeper look tomorrow but it cuts both the number of ops and some type assertion. Should definitely make iterations faster.

wdvxdr1123:
I realize my original implement (only when IteratorNext returns false, will the VM pop a iterator) was not correct when the loop body has break statements. So, I introduce a new opcode IteratorPop, and insert it behind the loop.

geseq:
I ran some naive benchmarks on the two implementations, and for 100,000,000 iterations, the difference in the two implementations was negligible, or generally within acceptable error margins for a benchmark like this. 

Example: `5.780540596s` vs `5.832257341s`

I'm not sure this justifies the change. @d5 thoughts?

d5:
Yeah. We should not risk the stability of the code if that's the case. Thanks for looking into this @geseq.

PR #347, opened 2021-09-05
For the OpMap VM instruction, since numElements is known just use the capacity hint.

deckarep:
This optimization perhaps is premature but I think it's a small easy win to just allocate the map precisely with the target number of elements since they are known from the bytecode operand.


geseq:
It’s unlikely to make any real difference irl as it’d presume defining a large map which in most cases would make the code unwieldy. But at the same time doesn’t hurt either

PR #349, opened 2021-11-06
Extract ModuleGetter interface for dynamic imports

### LINKED ISSUE #348: Dynamically generated code for import
Hi!

While using `tengo` embedded into some code, I wanted to make it possible to dynamically provide source code for imports, without having to import files from disk. Specifically, the import names I want to use do not map to files on disk, and are unknown before compiling a script (so I couldn't even somehow pre-generate them anywhere).

I couldn't find any easy way to do it, just a very ugly hack, based on trying to compile on a loop, handling any `CompilerError` on `ImportExpr` by generating the module I need, adding it to a `ModuleMap`, and trying again. Something like this:

```go
moduleMap := tengo.NewModuleMap()
script := tengo.NewScript(src)
script.SetImports(moduleMap)

var c *tengo.Compiled
var err error

done := false
for !done {
  done = true
  c, err = script.Compile()
  var cerr *tengo.CompilerError
  if errors.As(err, &cerr) {
    if node, ok := cerr.Node.(*parser.ImportExpr); ok {
      mSrc := generateModuleSource(node.ModuleName)
      moduleMap.AddSourceModule(node.ModuleName, mSrc)
      err = nil
      done = false
    }
  }
}

if err != nil { ... }

// use c here
```

But this is _beyond_ ugly.

There's a rather simple alternative: if we extract a trivial interface from `*tengo.ModuleMap` with just its `Get` method, and change a few places that declare parameters or fields with the new interface, I can pass an interface instead, and do my dynamic generation there.

E.g.:

```go
// (some changes omitted for brevity)

// tengo: modules.go:
type ModuleGetter interface {
  Get(name string) Importable
}

// tengo: script.go:
func (s *Script) SetImports(modules ModuleGetter) {
	s.modules = modules
}

// my_code.go:
type DynamicModuleGetter { ... }
func (d *DynamicModuleGetter) Get(name string) tengo.Importable {
  src := ...
  return &tengo.SourceModule{Src: src}
}
// ...
m := &DynamicModuleGetter{...}
script := tengo.NewScript(src)
script.SetImports(m)
c, err := script.Compile()
if err != nil { ... }
// use c here
```

This has the added benefit of getting the code closer to Go's good practice of taking interfaces instead of concrete types.

Will send a PR soon.

Thoughts?

bfreis:
Fixes #348.

Passes `make test`.

### DIFF
--- a/compiler.go
+++ b/compiler.go
@@ -49,7 +49,7 @@ type Compiler struct {
 	symbolTable     *SymbolTable
 	scopes          []compilationScope
 	scopeIndex      int
-	modules         *ModuleMap
+	modules         ModuleGetter
 	compiledModules map[string]*CompiledFunction
 	allowFileImport bool
 	loops           []*loop
@@ -63,7 +63,7 @@ func NewCompiler(
 	file *parser.SourceFile,
 	symbolTable *SymbolTable,
 	constants []Object,
-	modules *ModuleMap,
+	modules ModuleGetter,
 	trace io.Writer,
 ) *Compiler {
 	mainScope := compilationScope{
--- a/docs/interoperability.md
+++ b/docs/interoperability.md
@@ -168,6 +168,32 @@ mods.AddSourceModule("double", []byte(`export func(x) { return x * 2 }`))
 s.SetImports(mods)
 ```
 
+To dynamically load or generate code for imported modules, implement and
+provide a `tengo.ModuleGetter`.
+
+```golang
+type DynamicModules struct {
+  mods tengo.ModuleGetter
+  fallback func (name string) tengo.Importable
+}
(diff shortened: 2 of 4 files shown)


d5:
I like it. 

I think this can be useful for other use cases too. Can you update the [documentation](https://github.com/d5/tengo/blob/master/docs/interoperability.md#scriptsetimportsmodules-objectsmodulemap) in your PR?

d5:
@geseq what do you think?

geseq:
I agree. It’s a fairly simple change with little to no maintenance overhead but one that could be very useful.

geseq (approved):
LGTM but as @d5 mentioned would be good to add documentation

bfreis:
Good idea on the docs! Will add a commit with a simple example in the interop doc page.

Issue #332, opened 2021-05-08
QUESTION: file extension for Tengo script


damiva:
I would like to use Tengo for plugins scripting for my project.
Have you decided about file extension for Tengo script?
If no, i have tow suggestions:
**.tng** - consonant letters from **T**E**NG**O
**.xgo** - **X** is roman ten + **GO** (also it hints to GO lang)

Additionaly I would like to use IMPORT from files, BUT: can I restrict the use of files only within the selected directory?

geseq:
Tengo uses `.tengo` as the file extension for the files. It's descriptive and clear and I think better than 3 letter extensions that are hard to keep track of.

damiva:
thanks

Issue #286, opened 2020-05-22
Can tengo module support changing script suffix？


idoall:
vm.Compile() Error	{"error": "[uno.mshk]Compile Error: module file read error: open due.tengo: no such file or directory\n\tat (main):1:8"}

vm.Compile() Error	{"error": "[due.mshk]Compile Error: module file read error: open tre.tengo: no such file or directory\n\tat (main):1:8"}

ozanh:
Do you want to use another file extension instead of `.tengo` :scream: :open_mouth: ?  :-1:

idoall:
yes, file extension ".mshk"

idoall:
Tengo module file extension, can you make dynamic configuration?

d5:
Yeah, we should probably update the Compiler to take other extensions. `.tengo` can be appended only if the given module does not have any file extensions.

```golang
import("one") // one.tengo
import("one.tengo") // one.tengo
import("one.two") // one.two
```

ozanh:
What if somebody has `module.xyz.tengo` file already and imports it like `import("module.xyz")` ?

idoall:
> module.xyz.tengo

```
import("module.xyz.tengo")
```

d5:
Yeah, I think the general rule can be:

> `.tengo` is appended **only if** the name does not include any file extensions.

Users can still use the full file names if necessary.

ozanh:
I know that is an option @idoall but there are already running scripts and compiler needs to check file existence twice every time when path does not exist. I don't often see different extensions with python, js, etc.

d5:
We don't have to pre-check if the file exists or not. We simply check the file name and append `.tengo` if necessary but purely based on names. Maybe I'm missing something?

ozanh:
@d5 tengo cli runs bytecode and source file by checking extension. We need add a cli switch or bytecode file extension to distinguish different file types to run.

d5:
Aha. I forgot about it. But I guess this is for `import()` inside the script file. We can keep the current CLI logic to determine bytecode vs source file as is?

ozanh:
@d5 Let's wait for a week to see other feedbacks, if this is not urgent. I can create a PR but @idoall is this an urgent issue?

idoall:
It's OK to wait a week

@ozanh

ozanh:
@idoall It's been more than a week but my priorities has changed so my next PR will address this issue when I have time, sorry.

idoall:
[...]

PR #350, opened 2021-11-10
add custom extension support for importing source file

KEINOS:
This PR features issue #286 by implementing `tengo.Compiler.SetImportFileExt()` method which allows specifying custom extension names.

This feature aims for "`tengo`" as a library usage and should not affect the "/cmd/tengo" itself.

```diff
func main(){
...
	c := tengo.NewCompiler(srcFile, nil, nil, modules, nil)
	c.EnableFileImport(true)
	c.SetImportDir(filepath.Dir(pathFileSource))
+	err := c.SetImportFileExt(".tengo", ".mshk") // It will search *.tengo and *.mshk script files
...
}
```

- Related issue
  - #286
  - #332

---

- EDIT: Previously, in the first commit, `c.SetImportExt(".mshk")` searched for both `*.tengo` and `*.mshk` (`*.tengo` by default) with no input validation. <br>But currently it needs to explicitly define as `c.SetImportFileExt(".tengo", ".mshk")` for both file extensions. And returns an error if any given args are invalid as well.

### DIFF
--- a/compiler.go
+++ b/compiler.go
@@ -644,19 +644,15 @@ func (c *Compiler) SetImportDir(dir string) {
 // local module files.
 //
 // Use this method if you want other source file extension than ".tengo".
-// Note that this will replace the current list of extension name.
 //
 //     // this will search for *.tengo, *.foo, *.bar
 //     err := c.SetImportFileExt(".tengo", ".foo", ".bar")
 //
+// This function requires at least one argument, since it will replace the
+// current list of extension name.
 func (c *Compiler) SetImportFileExt(exts ...string) error {
-	if len(c.importFileExt) == 0 {
-		// At least one extension name is required.
-		c.importFileExt = []string{SourceFileExtDefault}
-	}
-
 	if len(exts) == 0 {
-		return nil // do nothing
+		return fmt.Errorf("missing arg: at least one argument is required")
 	}
 
 	for _, ext := range exts {
--- a/compiler.go
+++ b/compiler.go
@@ -640,10 +640,16 @@ func (c *Compiler) SetImportDir(dir string) {
 	c.importDir = dir
 }
 
-// SetImportExt sets the custom extension name of the source file for
+// SetImportFileExt sets the custom extension name of the source file for
 // loading local module files.
-func (c *Compiler) SetImportExt(ext string) {
+func (c *Compiler) SetImportFileExt(ext string) {
 	c.importExt = append(c.importExt, ext)
+}
+
+// GetImportFileExt returns a slice of custom extension name of the source
+// file for loading local module files.
+func (c *Compiler) GetImportFileExt() []string {
+	return c.importExt
 }
 
 func (c *Compiler) compileAssign(
--- a/compiler.go
+++ b/compiler.go
@@ -640,10 +640,34 @@ func (c *Compiler) SetImportDir(dir string) {
 	c.importDir = dir
 }
 
-// SetImportFileExt sets the custom extension name of the source file for
-// loading local module files.
-func (c *Compiler) SetImportFileExt(ext string) {
-	c.importExt = append(c.importExt, ext)
+// SetImportFileExt sets the extension name of the source file for loading
+// local module files.
+//
+// Use this method if you want other source file extension than ".tengo".
+// Note that this will replace the current list of extension name.
+//
+//     // this will search for *.tengo, *.foo, *.bar
+//     err := c.SetImportFileExt(".tengo", ".foo", ".bar")
+//
+func (c *Compiler) SetImportFileExt(exts ...string) error {
+	if len(c.importExt) == 0 {
+		// At least one extension name is required.
+		c.importExt = []string{SourceFileExtDefault}
+	}
+
+	if len(exts) == 0 {
+		return nil // do nothing
+	}
+
+	for _, ext := range exts {
+		if ext != filepath.Ext(ext) || ext == "" {
+			return fmt.Errorf("invalid file extension: %s", ext)
+		}
+	}
+
+	c.importExt = exts // Replace the hole current extension list
+
+	return nil
 }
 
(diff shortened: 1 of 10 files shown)


d5:
Thanks for the PR. I will have to look at the code later tonight, but can we change the function signature to something like:

```golang
SetImportFileExt(...string)
```

and it should replace the entire array instead of appending.

compiler.go
```diff
@@ -640,6 +640,12 @@ func (c *Compiler) SetImportDir(dir string) {
 	c.importDir = dir
 }
 
+// SetImportExt sets the custom extension name of the source file for
+// loading local module files.
+func (c *Compiler) SetImportExt(ext string) {
```
geseq, line 645: Maybe some basic validations here to make sure the extension isn't empty and is an actual extension.

geseq (changes requested):

KEINOS:
> @d5 
> can we change the function signature

ok. 
```diff
- func (c *Compiler) SetImportExt(ext string) {
+ func (c *Compiler) SetImportFileExt(exts ...string) {
```
> it should replace the entire array instead of appending.

roger roger. (￣^￣)ゞ

compiler.go
KEINOS: Indeed! I will add the basic validation to the method and test for it.

KEINOS: I added a simple validation. It now returns an error if any given extension is not valid.

```diff
- func (c *Compiler) SetImportExt(ext string) {
+ func (c *Compiler) SetImportFileExt(exts ...string) error {
````

Also, I changed the `Compiler`'s `importExt` field to `importFileExt`. Due to make it clear it is an extension for the source file and not an extension of tengo, such as a plug-in.

**Review by geseq — APPROVED** [review]

File: compiler.go [hunk 2]
```diff
@@ -640,6 +640,42 @@ func (c *Compiler) SetImportDir(dir string) {
...
+//
+func (c *Compiler) SetImportFileExt(exts ...string) error {
+	if len(c.importFileExt) == 0 {
+		// At least one extension name is required.
+		c.importFileExt = []string{SourceFileExtDefault}
+	}
+
+	if len(exts) == 0 {
+		return nil // do nothing
+	}
```
d5, line 660: why are we resetting `c.importFileExt` first? I think this function should just fail and return an error if `len(ext) == 0` (without doing anything).

KEINOS: > why are we resetting `c.importFileExt` first? 

You got the point.

I thought, in the first place, was for backward compatibility. In case the `Compiler` object wasn't created via `tengo.NewCompiler()` constructor. Now I realize this is not the right place to initialize, and it should not do more than the function name explains. 

> I think this function should just fail and return an error if `len(ext) == 0` (without doing anything).

That means that the function must contain at least one arg on use. If that's so, shouldn't the syntax be like:

```diff
- func (c *Compiler) SetImportFileExt(exts ...string) error {
+ func (c *Compiler) SetImportFileExt(ext string , exts ...string) error {
```

So to let the user know at least one arg is a must. Or just documenting so will do enough? What do you think?

KEINOS: Committed ad168e1 and it seems enough explanation, doesn't it?

compiler_test.go
```diff
@@ -1222,6 +1225,91 @@ func() {
...
+
+	c := tengo.NewCompiler(file, nil, nil, modules, nil)
+	c.EnableFileImport(true)
+
+	require.Equal(t, []string{".tengo"}, c.GetImportFileExt(),
+		"newly created compiler object must contain the default extension")
+}
+
+func TestCompilerSetImportExt_extension_name_validation(t *testing.T) {
+	c := new(tengo.Compiler) // Instanciate a new compiler object with no initialization
```
d5, line 1273: typo: `Instanciate`

compiler.go
```diff
@@ -1287,6 +1320,28 @@ func (c *Compiler) printTrace(a ...interface{}) {
 	_, _ = fmt.Fprintln(c.trace, a...)
 }
 
+func (c *Compiler) getPathModule(moduleName string) (pathFile string, err error) {
```
d5, line 1323: I know it's not directly related to this particular PR, but now that we allow changing the module file extension, I wonder if we should allow specifying the extension explicitly in the code. For example.

```golang
mod := import("my_module.some_ext")
```

should work even if `.some_ext` is not in `importFileExt` slice. What do you think?

d5 (changes requested):

compiler.go
geseq: I’d be fine with that but personally I think it’s a bit clunky, and creates two ways of doing imports where I’d rather have one.

d5: ok. that's fair. let's not do that for now. also if we allow extensions in the code it kind of breaks backward compatibility now that I think about it.

compiler_test.go
KEINOS: fixed! thanks!

compiler.go
KEINOS: I fixed the failing test @ 2d816c3. Plz, re-run the workflow when you have time.

KEINOS:
I wonder if we have to document it but couldn't determine where. Is the godoc enough?

d5:
we don't have documentation for that particular function, but maybe we can add a small note somewhere under here https://github.com/d5/tengo/blob/master/docs/tutorial.md#modules?

KEINOS:
> maybe we can add a small note somewhere under [here](https://github.com/d5/tengo/blob/master/docs/tutorial.md#modules)

[Done](https://github.com/d5/tengo/blob/92f052f9eebe931e2d507c351afec71e9b468116/docs/tutorial.md#modules). How about it? Is it clear enough?

geseq:
LGTM. Looks much better overall

d5 (approved):

PR #356, opened 2021-12-14
Fixed/inc dec

macheal:
start:="1"
end:="10"
for i:=start;i<end;i++{
println(i)
}
-----
1
11
111
1111
11111
111111
...endless loop


macheal (commented):
1. define:log type
2. inc/dec: add conditions

Issue #358, opened 2022-01-15
Discrepancy between "enum" package documentation and its behavior


KEINOS:
I would like to PR this issue, but I wanted to ask which I should do between:
  1. Modify the documentation to match the behavior of the script.
  2. Modify the script as written in the document.

## TL; DR

- The [`filter` function of the `enum` module](https://github.com/d5/tengo/blob/4846cf524306f9dc064f94c0283e72923e756b41/stdlib/srcmod_enum.tengo#L74-L87) does not work as described in the [documentation](https://github.com/d5/tengo/blob/4846cf524306f9dc064f94c0283e72923e756b41/docs/stdlib-enum.md?plain=1#L24-L27).
- Effet: On `enum.filter(x)`, if `x` is a map, then it returns `<undefined>`.
- Cause: As of v2.10.0, the [stdlib/srcmod_enum.tengo:L79](https://github.com/d5/tengo/blob/4846cf524306f9dc064f94c0283e72923e756b41/stdlib/srcmod_enum.tengo#L78-L87) checks only if it's `is_array_like()` and not `is_enumerable()`.
    https://github.com/d5/tengo/blob/4846cf524306f9dc064f94c0283e72923e756b41/stdlib/srcmod_enum.tengo#L78-L87

## TS; DR

- The document says as below.
  > `filter(x, fn) => [object]`: iterates over elements of x, returning an array of all elements fn returns truthy for. fn is invoked with two arguments: key and value. the key is an int index if x is array. **key is a string key if x is map**. It returns undefined if x is not enumerable.

- Script to reproduce (Tengo v2.10.0)
```go
fmt := import("fmt")
enum := import("enum")

filter := func(k, v) {
    return v % 2 == 0
}

// Array Case
arr := [1, 2, 3, 4, 5, 6, 7]
filtered := enum.filter(arr, filter)

fmt.println("filter(array):", filtered)
fmt.printf("  type: %T\n", filtered)

// Map Case
map := {"one": 1, "two": 2, "three": 3, "four": 4}
filtered = enum.filter(map, filter)

fmt.println("filter(map):", filtered)
fmt.printf("  type: %T\n", filtered)

// Output:
// filter(array):[2, 4, 6]
//   type: array
// filter(map):
//   type: undefined
```

misiek08:
Do you want it to return filtered map, array of values or array of tuples? I think @d5 should decide, in JS there is no `filter` for `map`/`Object` type :)

Script:
```
enum := import("enum")
[...]

PR #359, opened 2022-01-22
fix documentation for enum.filter

misiek08:
Fixes https://github.com/d5/tengo/issues/358


KEINOS:
Cool! 👍 

I think we should fix the script comment as well.
https://github.com/d5/tengo/blob/09240a1c1965574c6bfd2a7304a32993d2e2f267/stdlib/srcmod_enum.tengo#L74-L77

misiek08:
Will do this tomorrow, because I need to run/regenerate go file with this source and I need PC to do this :)

misiek08:
Sorry, took much longer because of some unexpected additional work I had to do.

KEINOS (commented):
@d5 
This change seems to be OK but doesn't this file be updated via `go generate` by the maintainers?

KEINOS (approved):
+1 to the changes though.

d5:
I think this PR includes the output of `go generate` as well, right? @misiek08

misiek08:
I ran `go run gensrcmods.go`

KEINOS:
@misiek08

> I ran `go run gensrcmods.go`

Ah, I got it.

That file should be triggered via `go generate ./...` defined in [`stdlib/stdlib.go`'s header](https://github.com/d5/tengo/blob/09240a1c1965574c6bfd2a7304a32993d2e2f267/stdlib/stdlib.go#L3), I think.
They will be overwritten anyway, but the fewer files to review, the better, right?

@d5 

I think it's good to go. What do you think?

KEINOS (approved):

PR #364, opened 2022-02-07
Fix parsing oct and hex numbers

### LINKED ISSUE #363: All ints are parsed in base 10
Octal and hexadecimal numbers are partially supported (by scanner), but always parsed in base 10.

#### Expected:

```
fmt := import("fmt")

fmt.println(0123)  // 83
fmt.println(123)   // 123
fmt.println(0x123) // 291
```

#### Actual:

```
fmt := import("fmt")

fmt.println(0123)  // 123
fmt.println(123)   // 123
fmt.println(0x123) // 0
```

qzb:
Fixes #363

### DIFF
--- a/parser/parser.go
+++ b/parser/parser.go
@@ -375,7 +375,12 @@ func (p *Parser) parseOperand() Expr {
 	case token.Ident:
 		return p.parseIdent()
 	case token.Int:
-		v, _ := strconv.ParseInt(p.tokenLit, 10, 64)
+		v, err := strconv.ParseInt(p.tokenLit, 0, 64)
+		if err == strconv.ErrRange {
+			p.error(p.pos, "number out of range")
+		} else if err != nil {
+			p.error(p.pos, "invalid integer")
+		}
 		x := &IntLit{
 			Value:    v,
 			ValuePos: p.pos,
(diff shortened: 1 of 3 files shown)


geseq (approved):
LGTM

d5 (changes requested):
Thanks for the PR. When I tested the following code at the VM-level:

```golang
expectRun(t, `out = 1 + 010`, nil, 9)    // works
expectRun(t, `out = 1 + 0x10`, nil, 17)  // works
expectRun(t, `out = 1 + 0o10`, nil, 9)   // Parse Error: expected ';', found o10
expectRun(t, `out = 1 + 0b10`, nil, 3)   // Parse Error: expected ';', found b10
```

I think we should fix this too if we want to fully support the binary/octal numbers?

qzb:
@d5 I've rewritten numbers scanning from scratch. Instead of trying to scan only valid numbers, it scans for things resembling number and let `strconv` do the validation. It should be fully compatible with golang's int and float literals.

If you don't like those changes, I can revert them and add support only for literals you mentioned.

geseq (approved):
Nice one!

geseq:
@d5 will need your approval as well

d5 (approved):

PR #366, opened 2022-02-07
Fix copying user functions

### LINKED ISSUE #365: Missing user function name in some error messages
When user function defined in module returns `ErrWrongNumArguments` or `ErrInvalidArgumentType`, error message doesn't include name of the function.

#### Example

```go
package main

import (
	"fmt"

	"github.com/d5/tengo/v2"
)

func main() {
	m := tengo.NewModuleMap()
	m.AddBuiltinModule("mod", map[string]tengo.Object{
		"test": &tengo.UserFunction{
			Value: func(...tengo.Object) (tengo.Object, error) {
				return nil, tengo.ErrInvalidArgumentType{}
			},
			Name: "test",
		},
	})

	s := tengo.NewScript([]byte(`import("mod").test()`))
	s.SetImports(m)

	_, err := s.Run()
	fmt.Println(err) // Prints:
	                 // Runtime Error: wrong number of arguments in call to 'user-function:'
	                 //        at (main):1:1
}
```

qzb:
Fixes #365

### DIFF
--- a/objects.go
+++ b/objects.go
@@ -1593,7 +1593,7 @@ func (o *UserFunction) String() string {
 
 // Copy returns a copy of the type.
 func (o *UserFunction) Copy() Object {
-	return &UserFunction{Value: o.Value}
+	return &UserFunction{Value: o.Value, Name: o.Name}
 }
 
 // Equals returns true if the value of the type is equal to the value of


d5:
PR looks good!

I know this is not related to the issue, but can you remove `EncodingID` field of `UserFunction` too? We don't use it anymore.

PR #371, opened 2022-02-18
Function kwargs

moisespsena:
Hello! 

I implement kwargs/keyword arguments for functions and closures. But I'm not very satisfied with the VM performance difference.

Could anyone help me with that? 

Original benchmark:
```
fibonacci(35)
Result:  9227465
Go:      50.13513ms
Parser:  14.368µs
Compile: 71.829µs
VM:      2.694701467s
fibonacci(35) (tail-call #1)
Result:  9227465
Go:      67.91345ms
Parser:  18.285µs
Compile: 68.48µs
VM:      2.776874189s
fibonacci(35) (tail-call #2)
Result:  9227465
Go:      236ns
Parser:  16.982µs
Compile: 55.363µs
VM:      18.95µs
```

Benchmark after my implementations:

```
fibonacci(35)
Result:  9227465
Go:      50.351214ms
Parser:  14.591µs
Compile: 71.252µs
VM:      14.926579001s
fibonacci(35) (tail-call #1)
Result:  9227465
Go:      74.248173ms
Parser:  19.069µs
Compile: 85.272µs
VM:      19.496196464s
fibonacci(35) (tail-call #2)
Result:  9227465
Go:      234ns
Parser:  17.122µs
Compile: 56.115µs
VM:      33.158µs
```


vm.go
```diff
@@ -536,79 +556,174 @@ func (v *VM) run() {
...
+				switch arr := v.stack[start+numArgs].(type) {
 				case *Array:
-					for _, item := range arr.Value {
-						v.stack[v.sp] = item
-						v.sp++
-					}
-					numArgs += len(arr.Value) - 1
+					newArgs := make([]Object, numArgs+len(arr.Value))
+					copy(newArgs, args)
+					copy(newArgs[numArgs:], arr.Value)
```
geseq, line 591: Same as above

vm.go
```diff
@@ -536,79 +556,174 @@ func (v *VM) run() {
...
 			if !value.CanCall() {
 				v.err = fmt.Errorf("not callable: %s", value.TypeName())
 				return
 			}
 
-			if spread == 1 {
-				v.sp--
-				switch arr := v.stack[v.sp].(type) {
+			for i := 0; i < numArgs; i++ {
+				args = append(args, v.stack[start+i])
```
geseq, line 583: This is probably playing a major part. Don't allocate new memory if you can work with existing. If you absolutely need this one create it with the size you need.

vm.go
```diff
@@ -536,79 +556,174 @@ func (v *VM) run() {
...
-			numArgs := int(v.curInsts[v.ip+1])
-			spread := int(v.curInsts[v.ip+2])
-			v.ip += 2
+			var (
+				numArgs           = int(v.curInsts[v.ip+1])
+				hasVarArgs        = int(v.curInsts[v.ip+2])
+				numKws            = int(v.curInsts[v.ip+3])
+				hasVarKw          = int(v.curInsts[v.ip+4])
+				args              []Object
+				kwargs, varKwargs map[string]Object
```
geseq, line 565: try to refactor to get rid of these

geseq: or allocate with the required sizes

geseq (commented):
In general you'll have to reduce the amount of type inference in your implementation. I suspect it's also playing a big part in the performance hit you're seeing.

vm.go
geseq: (below)

vm.go
```diff
@@ -536,79 +556,174 @@ func (v *VM) run() {
...
+			}
+
+			if callee, ok := value.(*CompiledFunction); ok {
+				calleeData := &ImmutableMap{
+					Value: map[string]Object{
+						"fn":     callee,
+						"args":   &ImmutableArray{Value: args},
+						"kwargs": &ImmutableMap{Value: kwargs},
+					},
+				}
```
geseq, line 637: Also try to see if you can avoid recreating this and update the existing one instead

geseq:
Hopefully these comments will help with a good starting point to optimize the runtime

geseq:
Much better but still quite slow. From a cursory glance not sure if anything else could be improved. Perhaps try profiling the benchmark code to see what the pain points are?

vm.go
```diff
@@ -536,65 +563,158 @@ func (v *VM) run() {
...
 		case parser.OpCall:
-			numArgs := int(v.curInsts[v.ip+1])
-			spread := int(v.curInsts[v.ip+2])
-			v.ip += 2
+			var (
+				numArgs           = int(v.curInsts[v.ip+1])
+				hasVarArgs        = int(v.curInsts[v.ip+2])
+				numKws            = int(v.curInsts[v.ip+3])
+				hasVarKw          = int(v.curInsts[v.ip+4])
+				kwargs, varKwargs map[string]Object
```
moisespsena, line 571: I believe that the declarations of the variables influence the performance, but are needed. I do not know how to improve this.

moisespsena, line 571: I believe that the declarations of the variables influence the performance, but are needed. I do not know how to improve this.

moisespsena, line 571: I believe that the declarations  of the variables influence the performance, but are needed. I do not know how to improve this.
```golang
var ( // ...
    kwargs, varKwargs map[string]Object
)
```

geseq: are you able to allocate it with the required capacity and then manipulate the contents as necessary? that might help

moisespsena: I removed the variable "Varkwargs" from the main scope, but I had no change in the benchmark.

I defined a capacity for the "kwargs" variable, but it worsened.

Still, I can not find a solution.

PR #374, opened 2022-04-13
Extended times lib functionality

vereecw:
Hello!

I'v taken the liberty to add some extra functions to the times library, namely:
- parseInLocation
- time_yearday
- time_in
- is_dst

All of these (besides parseInLocation) are 1 to 1 to their go counterpart.


docs/stdlib-times.md
```diff
@@ -72,6 +72,7 @@ times := import("times")
   2006" would be interpreted if it were the value; it serves as an example of
   the input format. The same interpretation will then be made to the input
   string.
+- `parseInLocation(format string, s string, loc string)`: parseInLocation is like parse but parse interprets a time as UTC; parseInLocation interprets the time as in the given location.
```
geseq, line 75: IIRC `parse` will allow the timezone for interpreting the time in the format string. If so this seems redundant.

docs/stdlib-times.md
```diff
@@ -114,6 +116,8 @@ times := import("times")
   t.
 - `time_string(t time) => string`: returns the time formatted using the format
   string "2006-01-02 15:04:05.999999999 -0700 MST".
+- `time_in(t time, loc string)`: returns a copy of t representing the same time instant, but with the copy's location information set to loc for display purposes.
```
geseq, line 119: maybe `to_location` instead, similar to `to_local` and `to_utc`

docs/stdlib-times.md
diericd: True, but if I'm not mistaken a location can have a different timezone per Location based on the current time.  At my location the timezone now is UTC+2 (CEST) but at 01/01/2022 it is UTC+1 (CET).

stdlib/times.go
```diff
@@ -555,6 +567,59 @@ func timesParse(args ...tengo.Object) (ret tengo.Object, err error) {
...
+		}
+		return
+	}
+
+	s3, ok := tengo.ToString(args[2])
+	if !ok {
+		err = tengo.ErrInvalidArgumentType{
+			Name:     "third",
+			Expected: "string(compatible)",
+			Found:    args[1].TypeName(),
```
prashant3286, line 601: Here, I think we have to use args[2] instead of args[1]

docs/stdlib-times.md
prashant3286, line 75: Here I think we can write all arguments as `parseInLocation(format, s, loc string)`

vereecw: True, but none of the other functions in this file are written like that

PR #375, opened 2022-04-15
Corrected/changed variable names of the base64 and hex imports

NAICOLAS:
Corrected/changed variable names of the base64 and hex imports.
Now corresponds to the module itself.


geseq (approved):

geseq:
Thanks for the PR! LGTM

PR #377, opened 2022-05-05
Reuse vm in compiled scripts

diericd:
Hi, 

I have a proposal here to re-use the VM in compiled tengo scripts. As it is now the creating of a VM every time a script.Run() is called has a signification overhead when it a called a lot. This is very noticeable when using a script to on every value of a incoming data stream.

### DIFF
--- a/script.go
+++ b/script.go
@@ -200,6 +200,7 @@ type Compiled struct {
 	globals       []Object
 	maxAllocs     int64
 	lock          sync.RWMutex
+	vm            *VM
 }
 
 // Run executes the compiled script in the virtual machine.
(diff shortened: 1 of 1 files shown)


geseq:
I assume you mean the compilation overhead. You shouldn't need to call `script.Run()` every single time, instead compile using `script.Compile()` and then you should then be able to run the the compiled byte code directly multiple times.

geseq:
To add to this, at any given point, creation of a `NewVM` would almost always be less expensive than the actual compilation, so I'm not even sure this change is all that useful to reduce the overhead.

geseq (commented):
I'm going to close this PR for now as I think it would open a can of worms to reuse the internal state of the VM this way. Please look at my prior comments for some ideas to reduce your runtime overhead

PR #381, opened 2022-06-11
update goreleaser to fix releases

geseq:
another one @d5. 

we have two options to fix releases, either increase compatibility to go 1.16+ or use newer goreleaser. more details here: https://github.com/goreleaser/goreleaser/pull/2069


d5 (approved):

geseq:
doesn't work, although v3 is supposed to have this fixed. will look into it

PR #387, opened 2022-07-10
fix runtime panics on self assignment

### LINKED ISSUE #386: Runtime panic on self assignment
The following Tengo script panics at run time:

a := a.foo

This likely because the compiler wrongly considers a to be defined after the := . A should only be considered defined in the next statement.

geseq:
fixes #386 

I'm not sure this is the best solution to the problem, but it's fairly minimal. I can't really think of anything other than functions that might need this recursive definition.

@d5 thoughts?

### DIFF
--- a/compiler.go
+++ b/compiler.go
@@ -692,12 +692,15 @@ func (c *Compiler) compileAssign(
 		return c.errorf(node, "operator ':=' not allowed with selector")
 	}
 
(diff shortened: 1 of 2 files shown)


geseq:
@d5 any chance you can review this and #388 soon, especially this one as it's a bugfix?

d5 (approved):

PR #389, opened 2022-07-28
Update README.md

harveer07:
grammar mistake fixed was missing an article


geseq:
Thanks for the PR. I wouldn’t claim to be an expert at this, but from googling I’m not entirely sure an article is required here. I will leave it for someone else to review.

richardwilkes:
The change looks correct to me.

richardwilkes (approved):

Issue #390, opened 2022-08-12
Compiler Question: why invert token.Less to token.Greater


shiyuge:
in compiler.go: https://github.com/d5/tengo/blob/6fc805399207b1afb7c59aab21dcab07c59b5df7/compiler.go#L144-L153

It seems that `<` is translated to `>` when compiling tengo scripts. I wonder why. This makes BinaryOp() in user defined types to never execute LessEq and Less branch.

For example I have a custom defined type `Number`, and 
1. `Number(14)>12` works because it would invoke `Number.BinaryOp` that I implemented
2. `Number(12)<14` throws a runtime error, because compiler inverts this to `14>Number(12)` in compilation process, and it would invoke `Int.BinaryOp` and fail.

Are there any special design considerations behind this? Can I simply send a PR to remove this piece of logic in compiler?

shiyuge:
I have opened a pull request for that:

https://github.com/d5/tengo/pull/391

PR #391, opened 2022-08-12
fix: do not invert token.Less to token.Greater in compiler

shiyuge:
I have posted an issue explaining this https://github.com/d5/tengo/issues/390


geseq (commented):
Thanks for the PR. Overall LGTM but can you please add a test similar to one you described in your issue for future regression testing?

shiyuge:
Thanks for the reply. A new test case is added

geseq (approved):

PR #392, opened 2022-08-30
fix: copy SourceMap in vm OpClosure branch

shiyuge:
It seems that SourceMap should be copied to the new CompiledFunction when executing OpClosure. All other fields are copied but this field is left out.

I've added a test to explain


geseq (approved):

geseq:
Thanks for the PR. LGTM

Issue #336, opened 2021-05-24
Some type of variables are not safely copied by Script.Clone.


matszpk:
My experiments with Script.Clone method give unexpected results. After cloning Compiled object, the global map and arrays values has been copied to origin Compiled object. Your code doesn't make new copy array values and map values, just simply copy globals array and your code of Clone doesn't make safe copy of Compiled object. Global mutable maps and arrays should be copied deeply to make safe-concurrent copy of state.

misiek08:
@matszpk can you provide simplest example or an test to show this behavior? This will be close to fixing so even patch is welcome :)

ipsusila:
First of all, thank you for the great works and contributions in Tengo. 
I just start to experiments with Tengo and looking at this issue. I thing the following example script shows what @matszpk means.

```go
// A script with two global variables assigned by host
// 1. `round` an integer type
// 2. `data` a map type

// 1. overwrite primitive type, safe
round = 1000
last_key := ""
for key, val in data {
    last_key = key
}

// 2. delete an entry from data (map type) is not safe in `cloned` bytecode
// the value in original compiled bytecode is being modified
delete(data, last_key)
```

I verified the behaviour, and propose a `patch` at [https://github.com/ipsusila/tengo/commit/14bfd45415b4c9861b1dfc5a894633dae43d3a92](https://github.com/ipsusila/tengo/commit/14bfd45415b4c9861b1dfc5a894633dae43d3a92).

geseq:
@ipsusila can you create a PR please

PR #393, opened 2022-09-22
Use Object.Copy when cloning globals in Compiled

ipsusila:
An attempt to fix issue #336

### DIFF
--- a/script_test.go
+++ b/script_test.go
@@ -608,3 +608,49 @@ func compiledIsDefined(
 ) {
 	require.Equal(t, expected, c.IsDefined(name))
 }
+
+func TestCompiled_Clone(t *testing.T) {
+	// simple script that modify a primitive and
+	// map-typed global variables
+	script := tengo.NewScript([]byte(`
+round = 1000			// overwrite primitive type
+last_key := ""
+for key, val in data {
+    last_key = key
+}
+delete(data, last_key)	// delete entry from data
+`))
+
+	params := map[string]interface{}{
+		"round": 1,
+		"data": map[string]interface{}{
+			"key1":    "test1",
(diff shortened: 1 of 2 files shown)


script_test.go
```diff
@@ -608,3 +608,49 @@ func compiledIsDefined(
...
+		require.NoError(t, err)
+	}
+
+	// retrieve values
+	round := compiled.Get("round").Int()
+	require.Equal(t, 1000, round) // round, modified to 1000 in script
+
+	data := compiled.Get("data").Map()
+	require.Equal(t, 3, len(data)) // data, one entry removed
+}
```
geseq, line 656: I suggest simplifying the test to keep it focused

```suggestion
func TestCompiled_Clone(t *testing.T) {
	script := tengo.NewScript([]byte(`
count = count + 1
data["b"] = 2
`))

	err := script.Add("data", map[string]interface{}{"a": 1})
	require.NoError(t, err)

	err = script.Add("count", 1000)
	require.NoError(t, err)

	compiled, err := script.Compile()
	require.NoError(t, err)

	clone := compiled.Clone()
	err = clone.RunContext(context.Background())
	require.NoError(t, err)

	count := compiled.Get("count").Int()
	require.Equal(t, 1000, count)

	data := compiled.Get("data").Map()
	require.Equal(t, 1, len(data))
}
```

geseq (changes requested):

geseq:
@ipsusila  thanks for the PR. Let's simplify the test but otherwise good to go

ipsusila:
Thank you @geseq I'm agree with you.

geseq (approved):

PR #394, opened 2022-09-25
fix slice low/high index shadowing variables in vm

ipsusila:
Variables `low` and `high` are being shadowed. Therefore, when slice index is invalid (not an `integer`), the error message does not gives *correct* invalid type hint, and always return `Runtime Error: invalid slice index type: int`

### DIFF
--- a/vm.go
+++ b/vm.go
@@ -376,8 +376,8 @@ func (v *VM) run() {
 
 			var lowIdx int64
 			if low != UndefinedValue {
-				if low, ok := low.(*Int); ok {
-					lowIdx = low.Value
+				if lowInt, ok := low.(*Int); ok {
+					lowIdx = lowInt.Value
 				} else {
 					v.err = fmt.Errorf("invalid slice index type: %s",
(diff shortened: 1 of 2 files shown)


geseq:
@ipsusila did you mean to add the encode/decode to this PR?

ipsusila:
@geseq I'm sorry that the PR is confusing. After I submitting the PR which is the fix for low/high variables, I add commit which includes Encode/Decode to `Compiled` to `forked` repo. I did't realised that the later commit also included in this PR. 

For Encode/Decode feature, I need your suggestions, wether it's acceptable to add such feature. If yes, should I separate the PR, or can we simply change the title and submit both PR?

geseq:
No worries at all. 

Ideally please move it to a separate PR. I’ll take a look at it later.

ipsusila:
Thank you @geseq I revert Encode/Decode related commit from PR. For Encode/decode, later I will submit the PR.

geseq (approved):

PR #395, opened 2022-09-28
Add Encode/Decode to Compiled object

ipsusila:
Rationale:

Compared to other scripting solutions for golang, `tengo` has a capability to execute `bytecode`. However:

1. Currently, `Encode/Decode` is implemented only in `Bytecode` and global variables are not saved. In order to run a script, it needs to be compiled first.
2. Because of (1), it's not possible to deliver/embed compiled script with host application, if it contains global variable(s).

### DIFF
--- a/script.go
+++ b/script.go
@@ -1,8 +1,11 @@
 package tengo
 
 import (
+	"bytes"
 	"context"
+	"encoding/gob"
 	"fmt"
+	"io"
 	"path/filepath"
 	"sync"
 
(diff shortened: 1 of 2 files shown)


geseq:
Thanks again for the PR @ipsusila 

This is an interesting idea but I have a few concerns here:
1. Byte code / object type changes between  versions of tengo might mean that this binary version might not decode correctly or when it does it might not run correctly. This would add an additional dimension to managing backwards compatibility that I’m not entirely sure I like.
2. It might be unwieldy to manage different versions of the same binary with only difference being the global values. Since these can be changed at any point and encoding can be done at any point, the only way to communicate the real state of the script to its user is outside of this encoding.
3. Compilation itself is very cheap at the moment and can be cloned in memory as necessary for multiple runs. Given that is it really worth the trouble dealing with the above?

PR #397, opened 2022-11-01
Better support for TZ locations in the times module

ganehag:
I've added a `location` parameter to times.date to allow for creating a timestamp in a specific location. The call now works the same as the equivalent Go function. This parameter is optional so as not to break existing code.

Along with this I've also added `times.in_location(t time, l string)`, which works similar to [func (t Time) In(loc *Location) Time](https://pkg.go.dev/time#Time.In)


geseq (approved):

NAICOLAS:
Nice!

PR #400, opened 2023-01-03
Downgrade exclusive lock to shared lock when cloning a script

sean-:
Allow multiple Go routines to concurrently `Clone()` a `Compiled` script.


geseq (approved):

geseq:
Thanks for the PR! LGTM

PR #401, opened 2023-01-06
Update objects.md to fix StringArray object injection example

keenskelly:
Get example code to working state for StringArray object injection Just calling "print()" produces no output. Importing fmt and calling println() works. The output was also producing "one, twothree" instead of "one, two, three" as noted in the comment. This example code seems to have diverged from the implementation current at the time that it was written. This code now builds, executes, and produces the expected output for me.

### DIFF
--- a/docs/objects.md
+++ b/docs/objects.md
@@ -297,9 +297,11 @@ to add `StringArray` to the script:
 ```golang
 // script that uses 'my_list'
 s := tengo.NewScript([]byte(`
-    print(my_list + "three")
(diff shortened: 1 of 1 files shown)


geseq (approved):

geseq:
Thanks for the PR.

geseq:
waiting for #407

geseq:
Can you rebase master please?

geseq:
Merged since this is just a documentation change

PR #410, opened 2023-02-09
fix: array appending and memory problems

macheal:


geseq:
Thanks for the PR. Can you add tests to expand the coverage?

PR #414, opened 2023-02-25
Update script.go

sin3degrees:
I don't want to always new Script, just change input is OK. So I add the func SetInput. I'd like to use ScriptPool instead of creating Script frequently.


geseq:
Other than go.mod the change itself looks ok to me but I’d wait for @d5 to pitch in as well in case I’m missing any subtle implications of this.

PR #422, opened 2023-04-23
update go version to fix goreleaser

geseq:


d5:
Thanks for taking care of this @geseq!

I know it’s only release and test actions, but can this somehow break backward compatibility for those who are using Tengo from 1.16 codebase?

geseq:
It probably won’t work with 1.16 if someone upgraded to recent versions due to the math constants but 1.17 would be fine.

1.16 is fairly old and has been end of life for a fairly long time now so I’m not really sure it’s worth worrying about.

d5:
Got it. Should we just start a new major version `v3` and change the minimum supported version to Go 1.18?

geseq:
I don’t mind either way. Historically we’ve just maintained compatibility with the last 3 major versions of Go.

Personally to me v3 seems a bit of an overkill since the changes are mostly bug fixes and library expansions and not really major changes to tengo itself

d5 (approved):
True. We didn't have any major changes in Tengo. Maybe I wanted to change the major version mostly because we're using semantic versioning - breaking changes should increase the major version. Let's merge this one, but maybe we should consider using different versioning scheme starting from `v3` in the future.

geseq:
Makes sense!

PR #423, opened 2023-06-23
Allow defining custom builtins

edison-moreland:
This is a feature I've found very helpful in my own project, SceneEngine.

For my use case, I've got custom user scripts being imported by a runtime script and I need to expose custom builtin functions to that userscript.
I tried creating the builtin as globals, but globals aren't exposed to libraries when they're imported.

### DIFF
--- a/cmd/bench/main.go
+++ b/cmd/bench/main.go
@@ -219,7 +219,7 @@ func runVM(
 
 	start := time.Now()
 
-	v := tengo.NewVM(bytecode, globals, -1)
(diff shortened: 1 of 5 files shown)


geseq (changes requested):
Thanks for the MR. I like the idea for this but I'm not keen on breaking the `NewVM` call for current users for what is likely to be a rarer use case. 

Can we instead introduce functional options to pass in the builtins. That way the code remains unbroken for all current users and will also allow future introductions of more options while remaining backwards compatible.

PR #424, opened 2023-06-25
fix json decode

geseq:


geseq:
@d5 can I get a review on this please?

d5:
Looks good. Just curious how stdlib implementation handles this. Have you looked at it?

geseq:
It looks like go stdlib uses the type provided by the variable into which it is unmarshaling. 

So if you specify an `int`, it’ll unmarshal as int when legal. If the type provided is `any` it does float64.

tengo has to figure out the correct type by itself.

d5 (approved):

PR #433, opened 2023-10-10
fix/jump_painc

wlxwlxwlx:
[OpJumpFalse, OpAndJump, OpOrJump, OpJump], these four instructions have been changed to use 4 bytes to avoid precision loss and panic when the number of instructions exceeds the maximum of 16 bits (65535)

### DIFF
--- a/instructions.go
+++ b/instructions.go
@@ -28,6 +28,12 @@ func MakeInstruction(opcode parser.Opcode, operands ...int) []byte {
 			n := uint16(o)
 			instruction[offset] = byte(n >> 8)
 			instruction[offset+1] = byte(n)
+		case 4:
+			n := uint32(o)
+			instruction[offset] = byte(n >> 24)
+			instruction[offset+1] = byte(n >> 16)
+			instruction[offset+2] = byte(n >> 8)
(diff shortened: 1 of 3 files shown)


geseq (approved):
LGTM

wlxwlxwlx:
Can you send me the test cases that failed? Let me see what the problem is.

geseq:
https://github.com/d5/tengo/actions/runs/7105633457/job/19343271620#step:7:8

wlxwlxwlx:
`If true {10}; 3333 ` An error was reported while executing this test case, and I tracked the code. On line 269 of compiler.go.

When [OpJumpFalse, OpAndJump, OpOrJump, OpJump] is changed to int32, each instruction becomes 4 bits (previously 2 bits), so when executing curPos:=len (c. currentInstructions()), curPos becomes 10 instead of 8. All subsequent test cases, as long as these 4 instructions are used, the value of curPos will be larger than before (the difference between the values of curPos in the fix/jump and master branches must be a multiple of 2). Therefore, if this pr is merged, the test cases need to be updated. I submitted an updated  PR, please take a look.
https://github.com/d5/tengo/pull/438

PR #436, opened 2023-12-05
Revert "[OpJumpFalse, OpAndJump, OpOrJump, OpJump], these four instru…

geseq:
…ctions have been changed to use 4 bytes to avoid precision loss and panic when the number of instructions exceeds the maximum of 16 bits (65535) (#433)"

This reverts commit 18424deb5a3a15277bc115fa173b9d27ac691583.


geseq:
@d5 can you approve this please.

d5 (approved):

PR #442, opened 2024-02-24
Throw a runtime error when trying to slice an unsupported type

mkuznets:
Add a missing switch case corresponding to a slice of an unsupported type. Without it Tengo continues execution past the unsupported slice with an invalid stack state, causing unexpected behaviour or obscure errors.

Example:

```golang
test := func() {
    a := 42
    return undefined[:5]
}

fmt := import("fmt")
fmt.println(test())  // -> prints 42
```

This PR changes it to throw a runtime error instead, which is consistent with similar cases such as using a function call operator on an unsupported type.

### DIFF
--- a/vm.go
+++ b/vm.go
@@ -534,6 +534,8 @@ func (v *VM) run() {
 				}
 				v.stack[v.sp] = val
 				v.sp++
+			default:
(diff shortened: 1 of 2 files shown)


geseq (approved):
Thanks for the PR. LGTM.

PR #446, opened 2024-04-23
feat: allow removal of specific attributes inside a builtin module

### LINKED ISSUE #445: Allow removal of specific attributes from builtin module
Similarly to https://github.com/d5/tengo/issues/319#issuecomment-749704118 I'd like to be able to use the "times" module from the stdlib but I don't want to allow script authors to call `times.sleep` and block the goroutine that is running the script for an arbitrarily long time. At the moment, the only way to do this that I can see is to copy/paste the contents of `times.go` to my own package and delete the sleep function from the file and module map, and then add that module instead of the stdlib one. Obviously this isn't a great solution.

I think it would be useful to allow removing specific attributes (e.g. functions) from a built in module via the `ModuleMap` e.g. by changing the `.Remove(name string)` method to be `.Remove(name string, attrNames ...string)`. This would be a non-breaking change to the API due to the use of variadic args for the `attrNames`.

lhopcroft123:
closes #445

### DIFF
--- a/modules.go
+++ b/modules.go
@@ -1,5 +1,7 @@
 package tengo
 
+import "fmt"
+
 // Importable interface represents importable module instance.
 type Importable interface {
 	// Import should return either an Object or module source code ([]byte).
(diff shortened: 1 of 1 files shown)


geseq:
Can we add a test case covering the new path please?

PR #454, opened 2025-04-10
Allow builtin modules replacement in compiled scripts

DenKoren:
Motivation:

Compiled scripts can be executed concurrently out of the box.
At the same time, custom builtin modules can be passed to tengo script out from go, providing additional functionalities to the script.

If such builtin module is not designed for concurrent use, it becomes a limitation for the compiled tengo script. In most cases it is possible to 'rule out' concurrency use limitation on module's side, but in most cases this is 'fixed' by new lock introductions, which makes parallel tengo compiled scripts execution to block on single instance of builtin module during runtime.

For now, the only way to overcome this issue is to re-compile template for each instance of builtin module, which works well but leads to code overcomplication: it becomes necessary to manage compiled script pools in order to execute each script instance solely single-threaded because it is strictly bound to single builtin module instance.

This feature allows to hot-swap builtin module instance inside compiled script making it possible to execute several clones (`.Clone()`) in parallel, each using its own instance of builtin module.

### DIFF
--- a/bytecode.go
+++ b/bytecode.go
@@ -16,6 +16,14 @@ type Bytecode struct {
 	Constants    []Object
 }
 
+func (b *Bytecode) Clone() *Bytecode {
+	return &Bytecode{
+		FileSet:      b.FileSet,
+		MainFunction: b.MainFunction,
+		Constants:    append([]Object{}, b.Constants...),
+	}
+}
+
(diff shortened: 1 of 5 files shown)


DenKoren:
Closed in favour of #473 : newer PR has also complex tests for parallel scripts execution.

PR #460, opened 2025-05-23
fix regex alternation

### LINKED ISSUE #459: Runtime error: slice bounds out of range in textREFind when using alternation patterns
## Description
When using the text library with a regex pattern containing multiple capture groups and alternation operator `|`, the library throws a runtime error. The error occurs specifically with patterns like `([a-zA-Z])|([0-9])`.

## Example Code
```
fmt := import("fmt")
text := import("text")
pattern := "([a-zA-Z])|([0-9])"
text_content := "a"
result := text.re_find(pattern, text_content, -1)
fmt.println(result)
```

## Error Message
runtime error: slice bounds out of range [:-1]

## Root Cause Analysis
The issue stems from the implementation of the `textREFind` function, which internally uses Go's standard library `regexp.FindAllStringSubmatchIndex`. When certain capture groups don't match, `FindAllStringSubmatchIndex` returns indices containing `-1`. These negative indices are then used in a slice operation `s2[-1:-1]`, causing the runtime error.

## Expected Behavior
The regex pattern should correctly match either letters or numbers as defined in the alternation groups.

## Actual Behavior
The program crashes with a slice bounds runtime error instead of performing the pattern matching.

## Environment
- Library version: [v2.17.0]
- Go version: [1.22.12]
- OS: [macOS]

## Additional Notes
This is a critical bug that affects pattern matching functionality when using alternation with capture groups. The issue is reproducible and affects core functionality of the text library.

geseq:
fixes #459

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -8,17 +8,17 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: check out
-        uses: actions/checkout@v2
-      - name: Unshallow
-        run: git fetch --prune --unshallow
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
       - name: set up Go
-        uses: actions/setup-go@v2
+        uses: actions/setup-go@v4
         with:
           go-version: 1.18
       - name: run goreleaser
-        uses: goreleaser/goreleaser-action@v3
+        uses: goreleaser/goreleaser-action@v5
         with:
           version: latest
-          args: release --rm-dist
+          args: release --clean
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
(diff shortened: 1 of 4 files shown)


geseq:
@d5 can you review please?

d5 (approved):

lxzyork:
@geseq Hi geseq, I'm having trouble importing the tengo package v3 in go.mod, getting the following error.
```
$ go get github.com/d5/tengo@3.0.0
go: errors parsing go.mod:
go.mod:10:2: require github.com/d5/tengo/v3: version "v3.0.0" invalid: go.mod has non-.../v3 module path "github.com/d5/tengo/v2" (and .../v3/go.mod does not exist) at revision v3.0.0
```
What steps should I take? Thank you for your guidance.

PR #465, opened 2025-06-05
func define allow at back

runsys:


parser/parser.go
```diff
@@ -148,6 +148,30 @@ func (p *Parser) ParseFile() (file *File, err error) {
 		return nil, p.errors.Err()
 	}
 
+	var nass [][]string
+	for i := 0; i < len(stmts); i += 1 {
+		nas := regexFindAll(stmts[i].String(), "[A-Za-z][a-zA-Z0-9]*(_[a-zA-Z0-9]+)*\\s*[:()]")
```
geseq, line 153: This parser logic based on regex looks quite brittle to me.

objects.go
```diff
@@ -298,7 +298,12 @@ func (o *Bool) IsFalsy() bool {
 // Equals returns true if the value of the type is equal to the value of
 // another object.
 func (o *Bool) Equals(x Object) bool {
-	return o == x
+	switch i:=x.(type) {
+	casse  *Bool:
```
geseq, line 302: Typo?

geseq (changes requested):

geseq:
Thanks for the PR. I think you’d have to rework this significantly and add thorough set of tests to consider this for merging.

PR #473, opened 2025-12-11
Replace builtin modules in compiled script

DenKoren:
Compiled code can be executed concurrently out of the box thanks to `Clone()` method of compiled script.

The problem is that custom builtin module, given to the script as a 'dependency' may not support this. Or it may require customisation for each clone to work properly (say, you provide a driver, transaction, connection or whatever else that should be unique for each script).

Inability to update/replace builtin modules inside script after compilation makes it impossible to re-use compiled script and requires compilation for each run.

This PR solves the mentioned limitation by extending compiled script's API.
The change is implemented maintaining backward compatibility with old behaviour, where bytecode was shared across several cloned instances of script.

Bytecode gets copied only when builtin modules need to be replaced after compilation (kind of 'copy-on-write' approach)

### DIFF
--- a/script.go
+++ b/script.go
@@ -244,6 +244,8 @@ func (c *Compiled) RunContext(ctx context.Context) (err error) {
 	return
 }
 
+// Size of compiled script in bytes
+// (as much as we can calculate it without reflection and black magic)
 func (c *Compiled) Size() int64 {
 	c.lock.RLock()
 	defer c.lock.RUnlock()
--- a/script.go
+++ b/script.go
@@ -268,6 +268,11 @@ func (c *Compiled) Clone() *Compiled {
 	return clone
 }
 
+// ReplaceBuiltinModule replaces a builtin module with a new one.
+// This is helpful for concurrent script execution, when builtin module does not support
+// concurrency and you need to provide custom module instance for each script clone.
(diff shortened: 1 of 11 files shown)


geseq:
Thanks for the PR. This looks like a good idea overall. On a high level this looks good to me, but will review deeper today/tomorrow. 

@d5 can you take a look as well please?

script.go
```diff
@@ -336,3 +361,24 @@ func (c *Compiled) Set(name string, value interface{}) error {
 	c.globals[idx] = obj
 	return nil
 }
+
+func (c *Compiled) prepareForModulesUpdate() {
```
geseq: I would just inline this into ReplaceBuiltinModule or lock the mutex here. Otherwise someone has to remember to lock correctly if they were reusing this function in the future

geseq (changes requested):
LGTM overall

script.go
DenKoren: Fixed.
Fetched fresh code from d5/master to resolve conflicts.

DenKoren: @geseq , sorry for long reaction, missed last comment :(

geseq (approved):

Issue #78, opened 2019-02-05
builtin function: 'freeze'


d5:
`freeze` to transform the value immutable recursively (deep immutability)

xkortex:
To clarify, if `foobar` is a nested object composed of primitives, maps, and arrays `freeze(foobar)` would return the equivalent data structure with ImmutableArrayS and ImmutableMaps?

What is the ordering convention on maps/ImmutableMaps? Just insertion order?

PR #480, opened 2026-04-28
Add freeze() builtin for deep immutability (#78)

ganehag:
freeze(x) recursively converts Arrays to ImmutableArrays and Maps to ImmutableMaps. Already-immutable values with no mutable children are returned as the same pointer (zero allocation). A memo table prevents infinite recursion on self-referential structures.

Primitives, strings, bytes, functions and errors pass through unchanged.

Closes the planned item from docs/freeze-builtin.md.


geseq (approved):
LGTM. Thanks for the PR!

